PR #6, opened 2021-01-19
feat: implement ygg status

gravitypriest:
Signed-off-by: Jeremy Crafts <jcrafts@redhat.com>

### DIFF
--- a/cmd/ygg/main.go
+++ b/cmd/ygg/main.go
@@ -153,58 +153,14 @@ func main() {
 		},
 		{
 			Name:  "status",
-			Usage: "reports status of connection and activation",
-			Flags: []cli.Flag{
-				// TODO: load this value from config ? (or have a default)
-				&cli.StringSliceFlag{
-					Name: "broker",
-				},
-			},
+			Usage: "reports connection status",
 			Action: func(c *cli.Context) error {
-				hostname, err := os.Hostname()
+				s, err := getStatus()
 				if err != nil {
-					// default for messaging if hostname can't be read
-					hostname = "host"
+					return cli.NewExitError(err, 1)
 				}
-				fmt.Printf("Connection status for %s:\n\n", hostname)
+				fmt.Println(s)
 
-				// get RHSM registration status
-				uuid, err := getConsumerUUID()
-				if err != nil {
-					fmt.Printf("⛔️ ERROR: Unable to check Red Hat Subscription Manager status: %s\n\n", err)
-				} else {
-
-					if uuid != "" {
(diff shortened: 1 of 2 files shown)


cmd/ygg/main.go
```diff
@@ -154,8 +154,58 @@ func main() {
...
+
+					if uuid != "" {
+						fmt.Println("✅ Connected to Red Hat Subscription Manager (RHSM).")
+					} else {
+						fmt.Println("❌ Not connected to Red Hat Subscription Manager (RHSM).")
+					}
+				}
+
+				// get broker connection status
+				errMsg := "⛔️ ERROR: Unable to connect to Cloud Client service: %s\n\n"
```
subpop, line 185: Until we have a D-Bus API for the client program to communicate with the service over, I think I'd prefer to keep this just reporting whether or not the `yggd.service` is running. This implementation bleeds a lot of daemon features into the client.

cmd/ygg/main.go
```diff
@@ -154,8 +154,58 @@ func main() {
...
+						// connection error
+						fmt.Printf(errMsg, err)
+					} else {
+						fmt.Println("✅ Cloud Client service is active.\n")
+					}
+				}
+				fmt.Println("See all your connected systems: http://red.ht/connect\n")
+
+				// get Insights registration status
+				// TODO (?)
```
subpop, line 201: Let's not commit TODOs. They should exist as issues in an issue tracker.

cmd/ygg/main.go
```diff
@@ -154,8 +154,58 @@ func main() {
 		{
 			Name:  "status",
 			Usage: "reports status of connection and activation",
+			Flags: []cli.Flag{
+				// TODO: load this value from config ? (or have a default)
+				&cli.StringSliceFlag{
```
subpop, line 159: This won't be necessary if we are only reporting on the status of the systemd service.

cmd/ygg/main.go
```diff
@@ -154,8 +154,58 @@ func main() {
...
+					hostname = "host"
+				}
+				fmt.Printf("Connection status for %s:\n\n", hostname)
+
+				// get RHSM registration status
+				uuid, err := getConsumerUUID()
+				if err != nil {
+					fmt.Printf("⛔️ ERROR: Unable to check Red Hat Subscription Manager status: %s\n\n", err)
+				} else {
+
```
subpop, line 176: Did you run this through `go fmt`? I thought it removed blank lines in places like this.

subpop (changes requested):

PR #13, opened 2021-02-17
build: install /usr/libexec/$LONGNAME

### LINKED ISSUE #11: install doesn't create /usr/libexec/rhc but it's required
For some reason, a normal RHEL instance has an un-owned `/usr/libexec/rhc` directory (`rpm -qf` doesn't give anything). Testing on an ostree instance, starting the rhcd daemon fails because that directory isn't there and can't even be created (ofc, it's RO).

```
			if localErr := processManager.BootstrapWorkers(p); localErr != nil {
				err = localErr
				quit <- syscall.SIGTERM
			}

			if localErr := processManager.WatchForProcesses(p); localErr != nil {
				err = localErr
				quit <- syscall.SIGTERM
			}
```

Both the above call play with `p := filepath.Join(yggdrasil.LibexecDir, yggdrasil.LongName)` - removing and reinstalling the package doesn't indeed create that directory but it's necessary to start the daemon.

runcom:
Close #11

cc @subpop

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,6 +129,7 @@ install: $(BINS) $(DATA)
 	install -D -m644 ./ygg.bash $(DESTDIR)$(DATADIR)/bash-completion/completions/$(SHORTNAME)
 	install -D -m644 ./yggd.bash $(DESTDIR)$(DATADIR)/bash-completion/completions/$(SHORTNAME)d
 	install -D -m644 ./data/pkgconfig/yggdrasil.pc $(DESTDIR)$(PREFIX)/share/pkgconfig/$(LONGNAME).pc
+	install -d -m755 $(DESTDIR)$(LIBEXECDIR)/$(LONGNAME)
 
 .PHONY: uninstall
 uninstall:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -129,6 +129,7 @@ install: $(BINS) $(DATA)
 	install -D -m644 ./ygg.bash $(DESTDIR)$(DATADIR)/bash-completion/completions/$(SHORTNAME)
 	install -D -m644 ./yggd.bash $(DESTDIR)$(DATADIR)/bash-completion/completions/$(SHORTNAME)d
 	install -D -m644 ./data/pkgconfig/yggdrasil.pc $(DESTDIR)$(PREFIX)/share/pkgconfig/$(LONGNAME).pc
+	mkdir $(DESTDIR)$(LIBEXECDIR)/$(LONGNAME)
```
subpop, line 132: Use `install -d` rather than `mkdir`. If for some terrifying reason `$(DESTDIR)` or `$(LIBEXECDIR)` or `$(LONGNAME)` don't exist by the time this `mkdir` is invoked, it will error. `install -d` will create the specified directory, and all components leading up to it.

subpop (changes requested):
Thanks for the PR!

Makefile
runcom: fixed (I guess that was me being extra lazy)

subpop (approved):

PR #16, opened 2021-03-11
register: use on-disk cert to get UUID

runcom:
Signed-off-by: Antonio Murdaca <runcom@linux.com>

### DIFF
--- a/canonical_facts.go
+++ b/canonical_facts.go
@@ -125,7 +125,7 @@ func GetCanonicalFacts() (*CanonicalFacts, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


subpop:
Ah yes. This. Currently, `ygg` has D-Bus dependencies on `com.redhat.RHSM1`. That requirement isn't going to be able to "go away" until we can figure out the registration backend feature. Because of that requirement, I didn't bother removing the GetConsumerUuid call in favor of simply reading the certificate on disk. I did remove all D-Bus dependencies from `yggd`, but not `ygg`.

subpop:
This isn't going to be our approach to fixing this dependency.

PR #28, opened 2021-04-15
BIOSUUID can be the empty string if sysfs doesn't provide one

maxamillion:
Signed-off-by: Adam Miller <admiller@redhat.com>

### DIFF
--- a/canonical_facts.go
+++ b/canonical_facts.go
@@ -120,9 +120,12 @@ func GetCanonicalFacts() (*CanonicalFacts, error) {
 		return nil, err
 	}
 
-	facts.BIOSUUID, err = readFile("/sys/devices/virtual/dmi/id/product_uuid")
-	if err != nil {
-		return nil, err
+	if _, err := os.Stat("/sys/devices/virtual/dmi/id/product_uuid"); !os.IsNotExist(err) {
+		BIOSUUID, err := readFile("/sys/devices/virtual/dmi/id/product_uuid")
+		if err != nil {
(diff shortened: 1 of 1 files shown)


subpop (changes requested):

canonical_facts.go
```diff
@@ -120,9 +120,14 @@ func GetCanonicalFacts() (*CanonicalFacts, error) {
...
-	if err != nil {
-		return nil, err
+	if _, err := os.Stat("/sys/devices/virtual/dmi/id/product_uuid"); !os.IsNotExist(err) {
+		BIOSUUID, err := readFile("/sys/devices/virtual/dmi/id/product_uuid")
+		if err != nil {
+			return nil, err
+		}
+		facts.BIOSUUID = BIOSUUID
+	} else {
+		facts.BIOSUUID = ""
```
subpop, line 130: Fun Go tidbit here. The `else` clause is unnecessary. Go initializes all variables to their "zero" value by default, so `facts.BIOSUUID` is already the zero value of `string` (which is an empty string `""`).

subpop (approved):

subpop (changes requested):
Commits need to be amended to follow the Conventional Commits pattern.

canonical_facts.go
maxamillion: I knew Go put all values to the zero value but I completely blanked when I was writing this... still learning! Thanks!

subpop (approved):

PR #32, opened 2021-09-28
feat(yggd): decouple MQTT client ID from certificate CN

subpop:
There are 3 new functions that perform operations on a client ID file:

1. `createClientID` will generate a semi-random string and set the contents of the client-id file to that string.
2. `getClientID` simply reads the contents of the file.
3. `setClientID` writes a value it is given to the file.

These 3 functions are using during the main program execution to determine a client ID value. The logic is as follows:

If a "cert-file" option is given, the certificate is parsed for a CN value. That value is then written to the client-id file. The client-id file is then read. If the contents of the read file is empty, a client-id is generated via `createClientID`. The global `ClientID` variable is then set to the value either read from the client-id file or generated if the value was empty.

The HTTP and MQTT client is configured with a customized TLS configuration. If no cert/key pair is given, the TLS configuration is left without a certificate for authentication. This lets `yggd` run without requiring a certificate and still connect to brokers anonymously.

### DIFF
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -159,10 +159,30 @@ func main() {
 			return cli.Exit(fmt.Errorf("cannot kill workers: %w", err), 1)
 		}
 
-		ClientID, err = parseCertCN(c.String("cert-file"))
-		if err != nil {
-			return cli.Exit(fmt.Errorf("cannot parse certificate: %w", err), 1)
+		clientIDFile := filepath.Join(yggdrasil.LocalstateDir, yggdrasil.LongName, "client-id")
+		if c.String("cert-file") != "" {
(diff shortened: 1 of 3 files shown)


lphiri:
@subpop  - were you reading my mind? I was going ask for this very thing :)

subpop:
This has been on my list of things to do for a while. I'm trying to find a middle-ground approach where the certificate CN value is used if it exists, otherwise it generates a client ID. Either way, it writes the value into a local state file (`/var/lib/yggdrasil/client-id`). I'm mulling over the ramifications of persisting the client ID in this way. For example, if a VM is cloned, the client ID comes with it. If the certificate CN value changes, it *does* read the new value from the certificate and write it to the state file, so that case should be okay.

PR #33, opened 2021-09-28
Conditionally configure certificate authentication

subpop:
Certificate authentication is only added to the TLS configuration if a cert/key pair are given. Otherwise, the HTTP requests will be unauthenticated. This is part of a larger effort to make certificates optional to the functionality of `yggd`.


subpop:
This changeset cannot be merged without #32, nor can #32 be merged without it. I'm closing the PR in favor of combining both into #32.

PR #37, opened 2021-12-06
Start workers explicitly

subpop:
This PR refactors the way processes are started, watched, killed, and restarted. The routines in `exec.go` were a mixture of higher-level worker process management (including restart delays, PID file management, etc.) and lower-level process start and kill routines. This PR moves the higher-level logic out of `exec.go` and into a file called `worker.go`. The routines in `exec.go` are now purely lower-level routines that start and stop processes; they have no awareness of worker configuration or process restart management (`startProcess`, `stopProcess`, `waitProcess`). The functions in `worker.go` operate on a `workerConfig` data structure to start, stop and manage workers.

A `workerConfig` structure is created by parsing a configuration file that must be defined in `/etc/yggdrasil/workers`. The existence of this file determines whether `yggd` is aware of, and will manage, a worker (previously this was handled by the existence of an executable program in `/usr/libexec/yggdrasil`). While a binary may still exist there, it will not be automatically started unless a valid configuration file points to that file as its executable. The name of the configuration file, minus the `.toml` suffix, is assumed to be the worker's handle (this ensures that no two workers will attempt to claim the same handle value).

For example, the echo worker now includes a configuration file:

```toml
exec="/usr/local/libexec/ygg-echo-worker"
protocol="grpc"
env=[]
```

The `exec` field must be a valid executable program. The `protocol` field is used to define which RPC protocol the worker is using. Currently the only supported value is "grpc". Values in the `env` array must be strings in the form of "VAR=VALUE" and are inserted into a worker's environment before starting it. Any value in the array that begins with either "PATH" or "YGG_" is ignored (in other words, a worker *may not* override its own `PATH` or any variable beginning with `YGG_` environment variables through its configuration file).

A worker's environment is automatically configured with a reasonable base PATH value (`/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`) and inherits values from its parent for: `http_proxy`, `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, and `NO_PROXY`.

A worker can be excluded from activation either permanently (by adding a value to the configuration file) or temporarily (by using the `--exclude-worker` command line flag). If a worker's name is found in this list during worker startup, it is skipped. It will not be started by `yggd` until the name is removed from the configuration value and `yggd` is restarted.

Little functionality should be changed by this PR. It mainly lays some groundwork to enable more functionality down the road. Nonetheless, this is a breaking change for workers, so it will not ship without advanced notice.

Fixes: ESSNTL-302, ESSNTL-298, ESSNTL-279

### DIFF
--- a/cmd/yggd/exec.go
+++ b/cmd/yggd/exec.go
@@ -1,119 +1,74 @@
 package main
 
 import (
-	"bufio"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"os"
 	"os/exec"
-	"path/filepath"
-	"time"
-
-	"git.sr.ht/~spc/go-log"
-	"github.com/redhatinsights/yggdrasil"
 )
 
-func startProcess(file string, env []string, delay time.Duration, died chan int) {
+// startProcess executes file, setting up the environment using the provided
+// env values. If started is not nil, it is invoked on a goroutine after the
+// process has been started.
+func startProcess(file string, args []string, env []string, started func(pid int, stdout io.ReadCloser, stderr io.ReadCloser)) error {
 	if _, err := os.Stat(file); os.IsNotExist(err) {
-		log.Warnf("cannot start worker: %v", err)
-		return
+		return fmt.Errorf("cannot find file: %v", err)
 	}
 
-	cmd := exec.Command(file)
+	cmd := exec.Command(file, args...)
 	cmd.Env = env
-
-	if delay < 0 {
-		log.Errorf("failed to start worker '%v' too many times", file)
-		return
-	}
-
-	if delay > 0 {
-		log.Tracef("delaying worker start for %v...", delay)
-		time.Sleep(delay)
-	}
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
-		log.Errorf("cannot connect to stdout: %v", err)
-		return
+		return fmt.Errorf("cannot connect to stdout: %v", err)
 	}
 
 	stderr, err := cmd.StderrPipe()
 	if err != nil {
-		log.Errorf("cannot connect to stderr: %v", err)
-		return
+		return fmt.Errorf("cannot connect to stderr: %v", err)
 	}
 
 	if err := cmd.Start(); err != nil {
-		log.Errorf("cannot start worker: %v: %v", file, err)
-		return
-	}
-	log.Debugf("started process: %v", cmd.Process.Pid)
-
-	go func() {
-		scanner := bufio.NewScanner(stdout)
-		for scanner.Scan() {
-			log.Tracef("[%v] %v", file, scanner.Text())
-		}
-		if err := scanner.Err(); err != nil {
-			log.Errorf("cannot read from stdout: %v", err)
-		}
-	}()
-
-	go func() {
-		scanner := bufio.NewScanner(stderr)
-		for scanner.Scan() {
-			log.Errorf("[%v] %v", file, scanner.Text())
-		}
-		if err := scanner.Err(); err != nil {
-			log.Errorf("cannot read from stderr: %v", err)
-		}
-	}()
-
-	pidDirPath := filepath.Join(yggdrasil.LocalstateDir, "run", yggdrasil.LongName, "workers")
-
-	if err := os.MkdirAll(pidDirPath, 0755); err != nil {
-		log.Errorf("cannot create directory: %v", err)
-		return
+		return fmt.Errorf("cannot start process: %v: %v", file, err)
 	}
 
-	if err := ioutil.WriteFile(filepath.Join(pidDirPath, filepath.Base(file)+".pid"), []byte(fmt.Sprintf("%v", cmd.Process.Pid)), 0644); err != nil {
-		log.Errorf("cannot write to file: %v", err)
-		return
+	if started != nil {
+		go started(cmd.Process.Pid, stdout, stderr)
 	}
 
-	go waitProcess(cmd, delay, died)
+	return nil
 }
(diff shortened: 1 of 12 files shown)


cmd/yggd/registry.go
```diff
@@ -0,0 +1,67 @@
...
+func (e registryError) Error() string { return string(e) }
+
+type registry struct {
+	mu sync.RWMutex
+	mp map[string]*worker
+}
+
+func (r *registry) set(handler string, w *worker) error {
+	if r.mp == nil {
+		r.mp = make(map[string]*worker)
```
eloycoto, line 20: should be locked here, no? Also the same for lines 37 and 48.

cmd/yggd/exec.go
```diff
@@ -1,191 +1,74 @@
...
 
-func killWorker(pidFile string) error {
-	data, err := ioutil.ReadFile(pidFile)
+// waitProcess finds a process with the given pid and waits for it to exit.
+// If died is not nil, it is invoked on a goroutine when the process exits.
+func waitProcess(pid int, died func(pid int, state *os.ProcessState)) error {
+	process, err := os.FindProcess(pid)
 	if err != nil {
-		return fmt.Errorf("cannot read contents of file: %w", err)
+		return fmt.Errorf("cannot find process with pid: %v", err)
```
eloycoto, line 47: In this function(startWorker) all the logs are well defined, but in some case like this we log the PID, in other the program, maybe will be good to add some metadata like log.WithFileds(programm, programme), so you know what worker failed, emit log, etc..

cmd/yggd/worker.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	env := []string{
+		"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+		"YGG_CONFIG_DIR=" + filepath.Join(yggdrasil.SysconfDir, yggdrasil.LongName),
+		"YGG_LOG_LEVEL=" + log.CurrentLevel().String(),
+		"YGG_CLIENT_ID=" + ClientID,
+		"http_proxy=" + os.Getenv("http_proxy"),
+		"HTTP_PROXY=" + os.Getenv("HTTP_PROXY"),
+		"HTTPS_PROXY=" + os.Getenv("HTTPS_PROXY"),
```
strider, line 66: Is there any reason to not support the lowercase https_proxy env var?

strider:
Ok, I've tested it again, but correctly this time. `yggd` detects new worker(s), starts them and stops them properly.

cmd/yggd/worker.go
subpop: I was modeling it off the [curl manpage](https://www.man7.org/linux/man-pages/man1/curl.1.html#ENVIRONMENT), but rereading that, I see that both upper and lower cases are supported, with the lower-case taking precedence. I'll refactor this to behave more like how `curl` is documented.

cmd/yggd/registry.go
subpop: Yep. Absolutely. Good catch.

cmd/yggd/exec.go
subpop: It looks like the logging could be improved overall, so I'll take a pass over the whole branch.

cmd/yggd/main.go
```diff
@@ -322,23 +333,28 @@ func main() {
...
-				go startProcess(filepath.Join(workerPath, info.Name()), env, 0, d.deadWorkers)
+			log.Debugf("starting worker: %v", info.Name())
+			config, err := loadWorkerConfig(filepath.Join(workerPath, info.Name()))
+			if err != nil {
+				log.Errorf("cannot load worker config: %v", err)
+				continue
 			}
+			if ExcludeWorkers[config.directive] {
+				continue
+			}
```
strider, line 345: It would be better to move this test before: 

`log.Debugf("starting worker: %v:", )`

Otherwise, we get a that debug message even if the worker has been excluded.

subpop: Yea that makes sense.

cmd/yggd/exec.go
subpop: I'm passing over the logging and error messages, and found a few improvements. Like here, it should include the PID as well as the underlying error returned from FindProcess. However at this layer, waitProcess knows nothing else about the program it is told to wait for. It's given a PID; it has no awareness of a worker or a worker config.

cmd/yggd/worker.go
```diff
@@ -0,0 +1,270 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"git.sr.ht/~spc/go-log"
+	"github.com/pelletier/go-toml"
+	"github.com/redhatinsights/yggdrasil"
+	"github.com/rjeczalik/notify"
+	"golang.org/x/net/http/httpproxy"
```
strider, line 19: Could you please tidy up the go.mod file as the golang.org/x/net is becoming a
direct dependency now?

strider:
Besides my comment about the `go.mod` file, it looks good to me and works pretty well.

subpop:
This one is starting to get out of control. I'm going to look into rebasing these commits down to a more atomic change.

subpop:
All wrapped up and ready to go. I'm pretty sure this is ready to merge now that the commits are more manageable.

strider:
Last force-push works as expected. Yes, I agree with your statement about merging
this big one as soon as possible. Nice work on that one, Link!

PR #39, opened 2021-12-09
Adding alternative transport: HTTP

jakub-dzon:
This PR introduces following changes:
 - transport abstraction layer to allow for implementing transports alternative for original MQTT;
 - HTTP transport support.


.gitignore
```diff
@@ -11,3 +11,5 @@
 /yggd.bash
 /yggdrasil.spec
 /doc/tags.toml
+
+.idea
```
subpop, line 15: This should not be part of a project-level gitignore file. This should be ignored globally in your `$XDG_CONFIG_HOME/git/ignore` file.

> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

subpop (changes requested):
This is a complex patch set. I suspect this will take a few iterations before we get it nailed down. This PR has prompted me to consider the project organization more concretely, so I deeply appreciate that. It has been a nagging, unanswered question for a long time; I am glad I finally have reason to force myself to find an answer.

Right off the top, I don't think I like the number of packages that gets created by this PR. I do want to move code from `cmd/yggd` into a supporting module level package eventually, but I don't think it's necessary to create this many deeply nested packages. Let's start with the following guidelines (which I'm adding to the CONTRIBUTING doc too).

* Code useful to `cmd/*` packages or external third-party packages should exist in the top-level `yggdrasil` package.
* Code useful to `cmd/*` packages, but not external packages, should exist in the top-level `internal` package.
* Code should exist in a package only if it can be useful when imported exclusively.
* Code can exist in a package if it provides an alternative interface to another package, and the two packages cannot be imported together.

Can we start by moving all the *.go files under `internal` into the `internal` package directly, rather than packages underneath `internal`? I think the fourth item in the list above might allow for a couple of packages under `internal` to exist in order to allow for the various concrete implementations of data types implementing the `Transport` interface.

jakub-dzon:
I've flattened the `internal` package structure only leaving `http` and `mqtt` sub-packages as two transport alternatives.

subpop:
Okay, I dove into this today and found myself experimenting a lot with the proposal. Overall, the idea is excellent, and the approach is sound. I wasn't quite happy with where API boundaries were drawn, so I spent a lot of time messing around with the Transport interface and the various methods of the two concrete implementations. I ended up making some fairly dramatic changes, but I'm happy with the overall outcome.

I moved the http-client.go into a package, `http`. This package offers a single type, `Client`, that acts as a specialized HTTP client configured with mutual TLS authentication. It lends itself nicely to being isolated into a package, as it is conceivable that this client could be used by external packages to send HTTP requests over an mTLS connection. (This package needs a good deal of improvement to actually provide a fully-featured HTTP client, but staging it in a package sets it up for future improvements.)

I decided to create an artificial API boundary in the Transport and Transport types where they only operate on byte slices. This creates a nice scope to draw around the transport layer, and ensures that they only operate on lower levels of data and not on yggdrasil Data or Control messages. Given this boundary, it is theoretically conceivable that the transport could be used to send and receive arbitrary data, so I ended up moving the Transport interface and the two transport types into a `transport` package.

This led me to discover a gap in the program that was previous filled by various "publish*" and "handle*" functions. I packaged all those up into a `Client` type that currently exists in `yggd.main`. This client is initialized with a `Transporter` and a `dispatcher` and mediates messages between the transporter and the dispatcher. It marshals data received by the transporter into messages and either sends them to the workers through the dispatcher or operates on them (in the case of control messages).

I pushed my changes to [a branch](https://github.com/RedHatInsights/yggdrasil/tree/http-transport-refactor), and you can see a diff between this branch and yours [here](https://github.com/jakub-dzon/yggdrasil/compare/upstreaming-http...RedHatInsights:http-transport-refactor). 

My review request for *you* is to take a look at my commit and let *me* know what *you* think. I definitely want this to be your PR still, but I found the easiest way to comment on the changes proposed was to move things around directly.

jakub-dzon:
Thank you for having a deeper look into this functionality and for the proposed changes. I like the way you pivoted the code around the transport layer and made it data model-agnostic. 
To make the code working I had to make a couple of [changes](https://github.com/jakub-dzon/yggdrasil/commit/49bb8b97234db16ed4fac1ff958a9558e751dcb2). 
What I would like to suggest is to change `Recv` in different method and type names to `Receive` (like [`RecvData`](https://github.com/jakub-dzon/yggdrasil/compare/upstreaming-http...RedHatInsights:http-transport-refactor#diff-3b3e1823cfec296bbe3b161adf96251de9c201becc4c59b2cf89ac787f01bfe8R12)).

subpop:
Yea, we can change `Recv` to `Receive`. I like the symmetry of 4 characters (like `RecvData` and `SendData`), but that's purely cosmetic, and I can understand the advantage to using the full English word. We've got code on 3 different branches now. Before things get any more out of hand, I'll push 92dd04bff6c86604c4ac01d09804bde155ffd225 to your branch here on this PR (I think I have commit access), then can you push your fixes and we'll keep moving forward on this PR.

jakub-dzon:
Yeah, let's do that.

subpop (approved):

subpop:
I'm happy with this. @jakub-dzon, are you?

jakub-dzon:
@subpop, thank you! I am happy with the final result as well. Could you, please, merge the changes?

PR #41, opened 2021-12-10
Extending the worker service to support graceful shutdown

jakub-dzon:
This PR introduces new worker service operation: `Disconnect` that can be called by yggdrasil when it receives Disconnect control message. Worker can use the new operation as a signal to execute graceful shutdown and clean-up.

The changes related to performing `Disconnect` operation call  will be provided as a separate PR, as they depend on #39.


jakub-dzon:
@subpop - could you, please, review?

subpop:
This is a reasonable extension to add to the protocol, though I think the name needs to be more explicit. "Disconnect": is it a command to the worker to disconnect (i.e. unregister itself), or is it a notification to the worker that a disconnection command was received, or is it a notification to a worker that an unexpected disconnection occurred.

To remove that ambiguity, I think this should get renamed to "DisconnectReceived". This is where an event notification system would be useful. Maybe we can prototype it here by making the method "NotifyEvent" with an event name parameter?

jakub-dzon:
I've changed the API to have a more general `NotifyEvent` operation instead of `Disconnect`. 
What do you think?

subpop:
Protobufs support enum types. I think it makes more sense to encode all possible event types into the protocol directly. Something like:

```
enum Event {
    // UNKNOWN is an undeclared event; protobuf requires the presence of a zero value. It shouldn't be used.
    UNKNOWN = 0;

    // RECEIVED_DISCONNECT is sent by the dispatcher to workers when the service received the "disconnect" command.
    RECEIVED_DISCONNECT = 1;

    // UNEXPECTED_DISCONNECT is sent by the dispatcher when the service unexpectedly disconnects from the network.
    UNEXPECTED_DISCONNECT = 2;
}

message EventNotification {
    Event name = 1;
    string detail = 2;
    map<string, string> metadata = 3;
}

rpc NotifyEvent(EventNotification) returns (EventReceipt) {}
```

jakub-dzon:
I thought about using enums, and I decided that would be too strict and any new event added to the protocol would require each worker needing it to update the dependency. Plain string representation allows for more flexible development.
If you think enum would still be better  I will change to enums.

subpop:
I think I'd prefer to keep event types more strictly typed as an enum. This focuses the change to events made by yggdrasil into a single location where all worker authors can see it. Workers will always need to have an error case for unsupported events, so new events shouldn't break their event handling logic, and protobuf messages are forward-compatible, so adding a new case to the enum shouldn't break the API with existing workers. Lastly, the `UNKNOWN` Event case, packaged into an `EventNotification` with a string value in `detail` is functionally equivalent to having a string as an `Event` type.

jakub-dzon:
OK, having documentation for free is a big plus here. I will change the code.

jakub-dzon:
Done. I added one more event definition that can be useful in pair with `UNEXPECTED_DISCONNECT` - `CONNECTION_RESTORED`.

subpop:
> Done. I added one more event definition that can be useful in pair with `UNEXPECTED_DISCONNECT` - `CONNECTION_RESTORED`.

Nice! Good thinking.

jakub-dzon:
@subpop Code sending the `DISCONNECT_RECEIVED` event is in #41 (which contains changes from this PR). What do you think about merging both API changes and the implementation at one go?

subpop:
> @subpop Code sending the `DISCONNECT_RECEIVED` event is in #41 (which contains changes from this PR). What do you think about merging both API changes and the implementation at one go?

Yea, I'm fine with both the protocol changes and the code implementing them existing in the same PR. I wanted to actually test this PR in isolation to make sure it doesn't break the existing dispatcher or workers. Including the implementation code in the same PR avoids that risk.

jakub-dzon:
OK. I will close this PR and leave just the #54 open.

PR #42, opened 2021-12-10
Additional environment variables

jakub-dzon:
This PR adds couple of environment variables that are passed to the workers:
 - `BASE_CONFIG_DIR` - yggdrasil configuration directory, within which the workers can create their own configuration subdirectory;
 - `LOG_LEVEL` - forwarding yggdrasil's log level to be applied by the workers;
 - `DEVICE_ID` - ID of the device as used by yggdrasil; to give workers possibility to use device identity in their logic.

### DIFF
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -343,9 +343,9 @@ func main() {
 		env := []string{
 			"YGG_SOCKET_ADDR=unix:" + c.String("socket-addr"),
 			"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
-			"BASE_CONFIG_DIR=" + configDir,
-			"LOG_LEVEL=" + level.String(),
-			"DEVICE_ID=" + ClientID,
+			"YGG_CONFIG_DIR=" + configDir,
+			"YGG_LOG_LEVEL=" + level.String(),
+			"YGG_CLIENT_ID=" + ClientID,
 		}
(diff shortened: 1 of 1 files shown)


cmd/yggd/main.go
```diff
@@ -339,10 +339,13 @@ func main() {
 		if err != nil {
 			return cli.Exit(fmt.Errorf("cannot read contents of directory: %w", err), 1)
 		}
-
+		configDir := filepath.Join(yggdrasil.SysconfDir, yggdrasil.LongName)
 		env := []string{
 			"YGG_SOCKET_ADDR=unix:" + c.String("socket-addr"),
 			"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+			"BASE_CONFIG_DIR=" + configDir,
```
subpop, line 346: Given the global namespace of environment variables, I think it's better to prefix these with `YGG_`, so `YGG_BASE_CONFIG_DIR` (or even just `YGG_CONFIG_DIR`), `YGG_LOG_LEVEL` and `YGG_DEVICE_ID`. `YGG_DEVICE_ID` would be more accurate if it where named `YGG_CLIENT_ID`.

subpop (changes requested):

cmd/yggd/main.go
jakub-dzon: Done.

subpop (approved):

PR #49, opened 2021-12-14
Improve the yggctl generate commands

subpop:
Previously, the yggctl generate commands relied on some quick marshalling and unmarshalling of map[string]interface{} types. This led to some bugs when trying to get JSON into the `content` field of messages. This PR updates the message generation logic to first create full yggdrasil message types, correctly populate the fields, and then marshal them into JSON before printing the message to standard output.

### DIFF
--- a/cmd/yggctl/generate.go
+++ b/cmd/yggctl/generate.go
@@ -9,13 +9,13 @@ import (
 	"github.com/redhatinsights/yggdrasil"
 )
 
-func generateDataMessage(messageType yggdrasil.MessageType, responseTo string, directive string, bytes []byte, metadata map[string]string, version int) (*yggdrasil.Data, error) {
+func generateDataMessage(messageType yggdrasil.MessageType, responseTo string, directive string, content []byte, metadata map[string]string, version int) (*yggdrasil.Data, error) {
 	var data interface{}
-	if err := json.Unmarshal(bytes, &data); err != nil {
+	if err := json.Unmarshal(content, &data); err != nil {
 		return nil, fmt.Errorf("cannot unmarshal content: %v", err)
 	}
 
-	content, err := json.Marshal(data)
+	body, err := json.Marshal(data)
 	if err != nil {
 		return nil, fmt.Errorf("cannot marshal data: %v", err)
 	}
@@ -28,7 +28,7 @@ func generateDataMessage(messageType yggdrasil.MessageType, responseTo string, d
 		Sent:       time.Now(),
 		Directive:  directive,
 		Metadata:   metadata,
-		Content:    content,
+		Content:    body,
 	}
 
 	return &msg, nil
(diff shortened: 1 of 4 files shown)


subpop:
FYI @strider @jakub-dzon

cmd/yggctl/generate.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/redhatinsights/yggdrasil"
+)
+
+func generateDataMessage(messageType yggdrasil.MessageType, responseTo string, directive string, bytes []byte, metadata map[string]string, version int) (*yggdrasil.Data, error) {
```
jakub-dzon, line 12: `bytes` parameter name is very generic; WDYT about `content` or something similar?

cmd/yggctl/generate.go
```diff
@@ -0,0 +1,102 @@
...
+			return nil, fmt.Errorf("cannot unmarshal control message: %v", err)
+		}
+		return &ctrl, nil
+	default:
+		return nil, fmt.Errorf("unsupported message type: %v", messageType)
+	}
+}
+
+// generateCommandMessage unmarshals bytes into a command message.
+func generateCommandMessage(messageType yggdrasil.MessageType, responseTo string, version int, bytes []byte) (*yggdrasil.Command, error) {
```
jakub-dzon, line 75: WDYT about changing `bytes` to something more descriptive, like `content`?

cmd/yggctl/generate.go
```diff
@@ -0,0 +1,102 @@
...
+		Metadata:   metadata,
+		Content:    content,
+	}
+
+	return &msg, nil
+}
+
+// generateControlMessage creates a control message of the appropriate type by
+// switching on the value of messageType.
+func generateControlMessage(messageType yggdrasil.MessageType, responseTo string, version int, bytes []byte) (*yggdrasil.Control, error) {
```
jakub-dzon, line 39: WDYT about changing `bytes` to something more descriptive, like `content`?

cmd/yggctl/generate.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if err := json.Unmarshal(bytes, &msg.Content); err != nil {
+		return nil, fmt.Errorf("cannot unmarshal content: %v", err)
+	}
+
+	return &msg, nil
+}
+
+func generateEventMessage(messageType yggdrasil.MessageType, responseTo string, version int, bytes []byte) (*yggdrasil.Event, error) {
```
jakub-dzon, line 91: WDYT about changing `bytes` to something more descriptive, like `content`?

cmd/yggctl/generate_test.go
```diff
@@ -0,0 +1,171 @@
...
+	messageType string
+	responseTo  string
+	directive   string
+	content     []byte
+	metadata    map[string]string
+	version     int
+}
+
+func TestGenerateDataMessage(t *testing.T) {
+	tests := []struct {
```
jakub-dzon, line 22: Have you considered using [Ginkgo](https://onsi.github.io/ginkgo/) and especially [table driven tests](https://pkg.go.dev/github.com/onsi/ginkgo/extensions/table?utm_source=godoc)? 
To me tests written with plain Go are pain to read.

cmd/yggctl/generate.go
subpop: This is funny. I had the parameter's originally named "content" and changed them to "bytes". I actually don't like `bytes` as much. It's the name of a stdlib package and can cause symbol collision later on if these files import the `bytes` package. I'll change them to `content`.

cmd/yggctl/generate_test.go
subpop: I have. I don't like taking on additional package dependencies without good reason, and adding a large module like Ginkgo as a dependency purely for testing isn't worth it. I also don't like the "language" of behavior-driven tests as much as table-driven tests; to me, they mask the actual code behind syntax sugar and make it more obtuse when trying to step-debug the tests. I find behavior-driven tests a pain to read. How about that!

jakub-dzon: :D

PR #54, opened 2021-12-20
NotifyEven worker service operation

jakub-dzon:
This PR introduces new worker service operation: `NotifyEvent` to signal occurrences of meaningful events within yggdrasil to workers. 

The PR provides support for `RECEIVED_DISCONNECT` event that is emitted to signal reception of a "disconnect" control message from the control plane. Workers can use it to execute graceful shutdown and clean-up.


cmd/yggd/client.go
```diff
@@ -185,3 +193,25 @@ func (c *Client) ConnectionStatus() (*yggdrasil.ConnectionStatus, error) {
 
 	return &msg, nil
 }
+
+func (c *Client) disconnectWorker(w worker) error {
```
subpop, line 197: There's a separation of responsibilities conflict here. The `yggd.Client` is a relatively new data structure introduced into `yggd`, so I don't think its role has been clearly defined yet. However, I noticed that this PR pulled the `protocol` package into `client.go` as an import; this caught my attention, and it occurred to me that perhaps the `yggd.Client` should not be aware of the underlying RPC protocol used to communicate with workers. This is particularly important, considering the future plans to support multiple RPC protocols.

This method (`disconnectWorker`) looks like it could easily be moved into `grpc.go` as a method on `dispatcher`. That feels more like an appropriate place for this RPC-related method to exist. And then the code block within the `CommandNameDisconnect` case statement can call `c.d.disconnectWorker`.

subpop (changes requested):

cmd/yggd/client.go
jakub-dzon: Good point; done.

subpop:
This works great. I had to enable the echo worker to get a log level value from the environment, since it was currently using the package default value exclusively. I also fixed a small typo in an error format string. I'm going to squash these commits though; I don't see any reason to keep the presence of the "Disconnect" method in the commit history; this feature should be an atomic change all wrapped up in one patch.

jakub-dzon:
Thank you very much for the reviews and for accepting the changes!

PR #58, opened 2022-01-18
TLS: Update config on certificate change

eloycoto:
In k4e-device-worker we need to rotate certificates quite often, but we
don't know how the user runs yggdrasil. This commit adds a certificate
inotify watcher, and update the clients if they changed.


internal/tls/tls.go
```diff
@@ -0,0 +1,146 @@
...
+	"k8s.io/utils/inotify"
+
+	"git.sr.ht/~spc/go-log"
+)
+
+const (
+	notifyEvents = unix.IN_MOVED_TO | unix.IN_CLOSE_WRITE
+)
+
+type TLSConfig struct {
```
subpop, line 20: Considering the idea in #59 where I suggested using a general purpose `main.Config`, I think this kind of logic could work very well within that config variable. It could hold a pointer to a `tls.Config`, run the inotify watchers to watch for changes to its file paths, and update either its own config values or the `tls.Config` variable with new data.

subpop (changes requested):
I think this could work really well if built on top of the global `main.Config` idea. Let's pursue that change first and then build this on top.

subpop:
Can we revisit this implementation by adding some inotify watch goroutines on the Config type?

eloycoto:
Closed this one

PR #59, opened 2022-01-18
Protobuf: Added dispatcher  getConfig method

eloycoto:
With this commit, workers are capable to retrieve yggd config, and be
able to do what it's needed at their end.

Fix: https://issues.redhat.com/browse/ESSNTL-306
Related-to: https://github.com/RedHatInsights/yggdrasil/discussions/51


cmd/yggd/grpc.go
```diff
@@ -34,9 +34,10 @@ type dispatcher struct {
 	workers     map[string]worker
 	pidHandlers map[int]string
 	httpClient  *http.Client
+	config      *pb.Config
 }
 
-func newDispatcher(httpClient *http.Client) *dispatcher {
+func newDispatcher(httpClient *http.Client, config *pb.Config) *dispatcher {
```
subpop, line 40: The dispatcher in this case would hold a pointer to a `main.Config` struct, and serialize it into a protobuf message within the `GetConfig` implementation function.

cmd/yggd/config.go
```diff
@@ -0,0 +1,19 @@
+package main
+
+import (
+	pb "github.com/redhatinsights/yggdrasil/protocol"
+	"github.com/urfave/cli/v2"
+)
+
+func getConfig(c *cli.Context) *pb.Config {
+	config := &pb.Config{
```
subpop, line 9: If we're going to store configuration state, I think I would rather abstract it inside a generic `main.Config` struct rather than directly into a protobuf message. protobuf isn't going to be the only wire format in use eventually, so this would set us up for supporting multiple RPC protocols.

cmd/yggd/main.go
```diff
@@ -166,25 +176,25 @@ func main() {
 			return cli.Exit(fmt.Errorf("cannot kill workers: %w", err), 1)
 		}
 
-		clientIDFile := filepath.Join(yggdrasil.LocalstateDir, yggdrasil.LongName, "client-id")
-		if c.String("cert-file") != "" {
-			CN, err := parseCertCN(c.String("cert-file"))
+		clientIDFile := filepath.Join(yggdrasil.LocalstateDir, yggdrasil.LongName, cliClientID)
+		if c.String(cliCertFile) != "" {
```
subpop, line 180: Again, if we're going to move to a global `main.Config` struct, this kind of initialization should probably be done from the values pulled out of the config variable. I can see some initialization code that puts together the config variable values from the `cli.Context`, while using the values from the config variable to further assemble the rest of the config variable.

```go
config = Config{}
config.ClientId = c.String("client-id")
config.CertFile = c.String("cert-file")

clientIDFile := filepath.Join(yggdrasil.LocalstateDir, yggdrasil.LongName, config.ClientID)
if config.CertFile != "" {
```

cmd/yggd/main.go
```diff
@@ -242,13 +252,13 @@ func main() {
 		switch c.String("protocol") {
 		case "mqtt":
 			var err error
-			transporter, err = transport.NewMQTTTransport(ClientID, c.String("server"), tlsConfig, client.DataReceiveHandlerFunc)
+			transporter, err = transport.NewMQTTTransport(ClientID, c.String(cliServer), tlsConfig, client.DataReceiveHandlerFunc)
```
subpop, line 255: If there's a global config variable, the transports can just get the server name from it instead.

protocol/yggdrasil.proto
```diff
@@ -98,3 +102,26 @@ message EventNotification {
     map<string, string> metadata = 3;
 }
 
+// A running-config definition for yggdrasil
+message Config {
+  // Log level
+  string ygg_log_level = 1;
```
subpop, line 108: This can just be `log_level`. There isn't a reason to prefix it with `ygg`, is there?

subpop (changes requested):
This is a nice approach. As I reviewed this, I came to the idea that we should probably store program configuration in a global `main.Config` struct; this sets up things for future RPC protocols that don't use protobuf, and solves some of the access to global configuration values that involve adding pointer fields to structures that need to know configuration. I think it's okay to grant global access to the configuration data to all data structures in the program.

subpop (changes requested):

cmd/yggd/grpc.go
```diff
@@ -34,9 +34,10 @@ type dispatcher struct {
 	workers     map[string]worker
 	pidHandlers map[int]string
 	httpClient  *http.Client
+	config      *Config
```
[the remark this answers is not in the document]
eloycoto: for dependency injection, I think that it's much better to be part of the dispatcher. IMHO

eloycoto:
@subpop I'll make some changes and I'll ping you when I finish it.

cmd/yggd/config.go
```diff
@@ -0,0 +1,59 @@
+package main
+
+import (
+	pb "github.com/redhatinsights/yggdrasil/protocol"
+	"github.com/urfave/cli/v2"
+)
+
+const (
+	cliLogLevel    = "log-level"
```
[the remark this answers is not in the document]
eloycoto: Already in the first commit, where only the config struct is added.

cmd/yggd/config.go
```diff
@@ -0,0 +1,59 @@
...
+	cliCaRoot      = "ca-root"
+	cliServer      = "server"
+	cliSocketAddr  = "socket-addr"
+	cliClientID    = "client-id"
+	cliTopicPrefix = "topic-prefix"
+	cliProtocol    = "protocol"
+	cliDataHost    = "data-host"
+)
+
+type Config struct {
```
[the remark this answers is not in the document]
eloycoto: All documented

eloycoto:
Hi @subpop  all your requested changes are now implemented and ready to review. 

Regards

subpop:
This is great. I noticed a few small changes, but I'll apply them and merge this.

subpop:
I rebased this on `main`, made a few minor edits and patched it up to work with the new start worker logic. One noticeable change I decided to make was to use a global `DefaultConfig` variable. The `startWorker` functions need to know things like socket address, so rather than passing a config value around, I opted for the global variable. Hopefully this decision will not come back to bite me later...

I also could not push changes to your PR branch, so I merged them onto `main` manually. Unfortunately, GitHub will never know that this PR was merged instead of closed.

PR #61, opened 2022-01-19
Mask the srpm for building as non-root

jordigilh:
Mask the `sprm` task to allow building as non-root by bypassing the `deps` target, which needs to be run as root. This change allows building `yggdrasil` in a container with a non-root user.

@subpop @jakub-dzon can you take a look?


subpop:
This spec isn't written with the intention of being used to build an RPM in a general case. This file exists explicitly to enable COPR builds. The `Makefile` at the root of the project is included as a utility to assist distribution packagers should they wish to package it. I have actually opened a [Review Request](https://bugzilla.redhat.com/show_bug.cgi?id=1988465) to package yggdrasil in Fedora, and I ended up not using the included `Makefile` at all (though with some recent changes to the `Makefile`, I probably could again).

I think I'd prefer to keep distribution-specific build instructions out of the upstream source tree as much as possible; I'd ideally remove the `.copr/Makefile` and `yggdrasil.spec` file to their own, COPR or Fedora-specific repository.

Issue #44, opened 2021-12-10
Backport exclude-worker options to 0.2


subpop:
The `--exclude-worker` configuration option is getting added to the main branch via #37, but this behavior might be desirable in the 0.2 line as well. Commit 12d4e86100a69cd5f5644a053fad2ec702f29240 is pretty well contained, and might be able to simply be cherry-picked onto `yggdrasil-0.2`. If not, it should get patched up and backported to `yggdrasil-0.2`.

PR #62, opened 2022-01-20
Backport-exclude-workers

subpop:
Add a `--exclude-worker` command flag and configuration parameter. Any strings listed in this argument will be stored in an exclusion list at run time. If a worker's *basename* matches a string in this list, it is excluded from execution when starting worker processes. For example, `--exclude-worker echo-worker` will prevent the file `/usr/libexec/yggdrasil/echo-worker` from being executed.

Fixes: #44


strider (approved):
Looks good to me.

PR #65, opened 2022-01-28
feat: HTTP transport add prefix

eloycoto:
Currently yggd supports MQTT prefix, in this way adds a prefix to the
MQTT topic. This commit adds the same feature for HTTP, so adds the
prefix to the http endpoint.

### DIFF
--- a/internal/transport/http.go
+++ b/internal/transport/http.go
@@ -3,10 +3,12 @@ package transport
 import (
 	"crypto/tls"
 	"fmt"
+	"path/filepath"
 	"sync/atomic"
 	"time"
 
 	"git.sr.ht/~spc/go-log"
+	"github.com/redhatinsights/yggdrasil"
 	"github.com/redhatinsights/yggdrasil/internal/http"
 )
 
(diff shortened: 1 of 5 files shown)


internal/transport/http.go
```diff
@@ -98,5 +100,6 @@ func (t *HTTP) send(message []byte, channel string) error {
 }
 
 func (t *HTTP) getUrl(direction string, channel string) string {
-	return fmt.Sprintf("http://%s/%s/%s/%s", t.server, channel, t.clientID, direction)
+	path := filepath.Join(yggdrasil.TopicPrefix, channel, t.clientID, direction)
```
jakub-dzon, line 103: How will that be used by Flotta?

eloycoto:
```
sudo ./yggd  \
  --log-level info \
  --cert-file /tmp/client.pem \
  --key-file /tmp/client.key \
  --protocol http \
  --topic-prefix /api/k4e-management/v1 \
  --ca-root /home/eloy/dev/upstream/eloycoto/mtls_example/certs/rootCA.pem \
  --server localhost:8443
```

jakub-dzon (approved):
LGTM

subpop (changes requested):
This looks great. Since we're already breaking the configuration file between 0.2 and 0.3, we should change this flag name to something more generic, so that it applies to both the MQTT and the HTTP transport.

* `--api-prefix`?
* `--path-prefix`?
* `--prefix`?

Any other suggestions?

eloycoto:
+1 to path-prefix

strider (approved):
Looks good to me.

PR #66, opened 2022-01-28
Feat: Added a way to select client-id source

eloycoto:
This commit adds a new way to select client-id source. By default it's
using client certificate CN. Using this way, a user can force the client
to be the machine-id, in this case will be a UUID.

The setup will look like this:

```
sudo ./yggd  \
  --log-level info \
  --cert-file /tmp/client.pem \
  --key-file /tmp/client.key \
  --protocol http \
  --topic-prefix eloy \
  --client-id-source machine-id \
  --ca-root /tmp/rootCA.pem \
  --server localhost:8443
```

Co-authored-by: Jakub Dzon <jdzon@redhat.com>


subpop (changes requested):
I think this adds unneeded complexity. Could we simply support a `--client-id` flag that, if present, overrides the value read from the `client-id` file? Your example command line could still work, replacing `--client-id-source machine-id` with `--client-id $(cat /etc/machine-id)`. And the existing logic of defaulting the client ID to a UUID derived from a certificate Subject could still exist.

subpop (approved):

PR #70, opened 2022-02-10
respect LDFLAG value from the environment

subpop:
If LDFLAGS is set in the environment before building, the flags in the
variable will get integrated into the LDFLAGS passed to the compiler.
This enables building with environment-dependent flags.

This is actually an upstreaming of a patch applied downstream in RHEL.

Resolves: RHBZ#2023489


subpop:
We should backport this to yggdrasil-0.2 too.

eloycoto (approved):

jakub-dzon (approved):

Issue #69, opened 2022-02-10
Invalid client ID causes MQTT transport to fail connection


subpop:
**Describe the bug**
Setting the client ID manually using `--client-id` permits any valid string character to be used as an MQTT client ID. This value is used as part of the topic strings when forming MQTT topic names, so characters such as `+` and `#` will result in topic names that are meant to be interpreted as wild-card topic values. The client refuses to connect in this case. In fact, [the MQTT specification](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901059) restricts the client ID topic character set to `0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ`.

**To Reproduce**
1. Run `yggd` with a manual client ID: `yggd --server tcp://test.mosquitto.org:1883 --client-id "test#"`

**Actual behavior**
The client exits with the following error:

```
cannot connect using transport: cannot connect to broker: network Error : EOF exit status 1
```

**Expected behavior**
The client should connect to the broker and subscribe to the topics.

subpop:
Fixed in https://github.com/RedHatInsights/yggdrasil/commit/6d13812c1e40f48a32ec9f97029475120c5b1d19

PR #72, opened 2022-02-11
Enforce value passed through the --client-id flag

strider:
The ClientID should contain only the following characters
'[a-zA-Z0-9]'. This patch checks the value passed through the command
line and fails gracefully if the value is not respecting the rule.

Fix: #69

Signed-off-by: Gael Chamoulaud (Strider) <gchamoul@redhat.com>

### DIFF
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"regexp"
 	"sync/atomic"
 	"syscall"
 	"time"
@@ -103,7 +104,7 @@ func main() {
 		},
 		&cli.StringFlag{
 			Name:  cliClientID,
-			Usage: "Use `VALUE` as the client ID when connecting",
+			Usage: "Use `VALUE` as the client ID when connecting. Should only contain the characters '[a-zA-Z0-9]'",
 		},
 		altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
 			Name:  cliExcludeWorker,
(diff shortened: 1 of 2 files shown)


cmd/yggd/main.go
```diff
@@ -204,6 +205,14 @@ func main() {
 			}
 		}
 
+		if DefaultConfig.ClientID != "" {
```
eloycoto, line 208: I'm ok with this, but I would move the validation to the config package, like :

```
if !DefaultConfig.IsValidClientID() { return cli.Exit(...)}
```

subpop:
This is interesting. This restriction only really applies to MQTT client IDs. It restricts the value of the `ClientID` field to only alphanumeric characters. This would disallow using the content of `/etc/machine-id` as a client ID. In fact, the ClientID for the HTTP transport should have a different set of allowable characters. A forward-slash or a question mark would break the URL path parsing.

By this reasoning, each transport should check the client-id for proper values and return an error.

cmd/yggd/main.go
```diff
@@ -143,17 +144,18 @@ func main() {
...
-			CertFile:       c.String(cliCertFile),
-			KeyFile:        c.String(cliKeyFile),
-			CaRoot:         c.String(cliCaRoot),
-			PathPrefix:     c.String(cliPathPrefix),
-			Protocol:       c.String(cliProtocol),
-			DataHost:       c.String(cliDataHost),
-			ExcludeWorkers: map[string]bool{},
+			LogLevel:        c.String(cliLogLevel),
+			ClientID:        c.String(cliClientID),
+			IsValidClientID: regexp.MustCompile(`^[a-zA-Z0-9]+$`).MatchString(c.String(cliClientID)),
```
eloycoto, line 149: this should be a method function not a property of the Config IMHO.

cmd/yggd/main.go
strider: @subpop @eloycoto 

Ok, I've moved this partially in the config package. But now, it means we disallow the content of `/etc/machine-id` as a client ID! :/

eloycoto:
> By this reasoning, each transport should check the client-id for proper values and return an error.

I do not think so, no? machine-id is a UUID so it's only numbers and letters, no? IMHO, the current regexp is a valid solution for us, and it's a reasonable approach.

subpop:
> I do not think so, no? machine-id is a UUID so it's only numbers and letters, no? IMHO, the current regexp is a valid solution for us, and it's a reasonable approach.

Oh you're right! I assumed it was a hexadecimal string format with hyphens (which I have been incorrectly referring to as "UUIDv4-formatted"). It appears to omit the hyphens.

Shouldn't the HTTP transport still refuse client ID strings that contain characters that will break URL parsing?

strider:
> > By this reasoning, each transport should check the client-id for proper values and return an error.
> 
> I do not think so, no? machine-id is a UUID so it's only numbers and letters, no? IMHO, the current regexp is a valid solution for us, and it's a reasonable approach.

Yes the `/etc/machine-id` is compliant but by default we create something like that:

```	
data := []byte(hostname + "-" + generatedUUID.String())

-> yggdrasil/test-client-test-vm-rhel90-gchamoul-4a705af3-8b87-11ec-a889-fa163ede1e0e/data/in
```

strider:
We actually don't use the `machine-id` content to generate the client ID if I am not wrong.

subpop:
> We actually don't use the `machine-id` content to generate the client ID if I am not wrong.

This is correct. The current logic is roughly as follows:

If `cert-file` is not empty, parse the contents of the file specified by the `cert-file` option and use the CommonName field from the certificate Subject.
If `cert-file` is empty, generate a client ID and write the value to /var/lib/yggdrasil/client-id. The generated client ID is the hostname + a UUID string, joined by a hyphen.

The generated string contains hyphens, so it would fail the validation. On hosts that use the RHSM consumer certificate as the value of `cert-file`, the `Subject.CommonName` value from the certificate is a hyphenated UUID string, so it would *also* fail validation.

subpop:
I actually don't like the implementation of adding a struct field to the `Config` type. So far, each struct field maps directly to a value in the flags slice (which in turn map to a value in the config file), and adding this field would break that behavior. I like how each struct field can be mapped to a flag value. The validation routine should probably be a method on the Config type. We also have to skip calling this validation routine when using the Subject's CN value as the client-id.

PR #73, opened 2022-02-15
Make cert-file and key-file required

subpop:
Assumptions are made by yggd 0.2.x that content will be available in the cert-file and key-file contents. When run without values for these flags, errors occur.


eloycoto (approved):

subpop:
Thank you both!

PR #74, opened 2022-02-25
Allow load client-id also from config file

machacekondra:
Fixes: https://github.com/RedHatInsights/yggdrasil/issues/75

Signed-off-by: Ondra Machacek <omachace@redhat.com>


machacekondra:
@subpop @eloycoto PTAL

eloycoto (approved):

eloycoto:
Commit fails @machacekondra change the subject to `fix: XXX`

subpop:
Nice catch! Thanks for the PR!

PR #77, opened 2022-03-22
fix: HTTP Transport return all responses

eloycoto:
I discovered that when the HTTP response is >400 the error is not nil,
and the transport didn't return the http.Response back to the worker.

With this commit, in case of the transport.Send fails, it returns an
error and also the http response to the worker that requested it.

Discovered on: https://github.com/project-flotta/flotta-operator/pull/144

### DIFF
--- a/internal/transport/http_test.go
+++ b/internal/transport/http_test.go
@@ -1,0 +1,177 @@
+// +build go1.16
+
+package transport_test
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	client "github.com/redhatinsights/yggdrasil/internal/http"
+	"github.com/redhatinsights/yggdrasil/internal/transport"
+)
+
+func startServer() *http.Server {
+	getResponse := func(w http.ResponseWriter) {
+		data := map[string]string{
+			"status": "OK",
+		}
+		resBytes, _ := json.Marshal(data)
+		fmt.Fprintf(w, "%s", resBytes)
+	}
+	port, err := getFreePort()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/transport/http_test.go
```diff
@@ -0,0 +1,141 @@
+// +build go1.16
```
eloycoto, line 1: This is mainly because http.Client in 1.15 is not able to have a nil TLS config, I do not want to over-complicate the thing because the version EOL, and the alternative is set certs, etc..

internal/transport/http_test.go
```diff
@@ -0,0 +1,141 @@
...
+		w.WriteHeader(http.StatusInternalServerError)
+		getResponse(w)
+	})
+
+	http.HandleFunc("/yggdrasil/test/200/out", func(w http.ResponseWriter, req *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		getResponse(w)
+	})
+
+	log.Fatalf("Cannot started mock http server: %v", http.ListenAndServe(":8889", nil))
```
jakub-dzon, line 42: `Cannot started` -> `Cannot start`

jakub-dzon, line 42: Please make sure that the server is shut down when this set of tests ends. Otherwise it will occupy chosen port until the process running tests finishes.

jakub-dzon: Also - it might be useful to use random port; there might be port clashes on dev machines, especially around 8*** range.

jakub-dzon (changes requested):

jakub-dzon (approved):

internal/transport/http_test.go
strider: Yes got it. I will modify the test matrix asap.

strider (approved):
Looks good to me.

Issue #45, opened 2021-12-12
Replace canonical-facts with a file


subpop:
`CanonicalFacts` is still a place where yggdrasil is dependent on insights-client; it looks for the file `/etc/insights-client/machine-id` and includes it in the facts. Rather than making the facts code look up values every time, convert it to a pattern much like client ID. If a file exists, say `/etc/yggdrasil/canonical-facts.json`, use that file as the facts data. This way we can drop the specifics of what makes up canonical facts about a host, making it opaque to `yggd`.

PR #83, opened 2022-05-10
fix: Using right cert on cannonicalFacts

eloycoto:
Because now cert-file can be set, the cannonicalFacts is still trying to
get the cert from invalid place, given an invalid message to the user.


subpop:
`CanonicalFacts` is quite fragile. I don't like how dependent it is on various system files and paths. I'd like to convert it to some structure that doesn't bind it so tightly to files like `/etc/insights-client/machine-id`. I filed this as #45. Would you be interested in taking this PR and further abstracting the `CanonicalFacts` away from its dependence on strict system paths?

eloycoto:
I'm happy to work on that, yes! But I'll love it if this can get merged as a middle step and a fix for us :-)

subpop (approved):

PR #85, opened 2022-08-09
Expose HOME to workers

adamruzicka:


subpop:
This is an interesting idea, but my first thought on how to implement a feature like this would be to create a separate home directory for each worker rather than re-exporting the `yggd` user's home directory. What is the feature or need that you're trying to fulfill by including HOME in the worker environment?

adamruzicka:
In Satellite, remote execution now has two backends. One uses ssh, where a connection is opened from Satellite to the target host and a script provided by the user is executed there. Due to how ssh handles things, the user-provided script is executed in home directory of the user that was used to open the connection.

The other backend is based on yggdrasil. There, due to systemd spawning yggdrasild in `/` and the worker process inheriting the working directory, the script provided by the user gets run in `/`. All I'm trying to accomplish here is to be consistent. Of course, we could make systemd set the working directory and everything would seamlessly inherit it, or just parse the home directory at workers startup from /etc/passwd, but this approach with exposing HOME felt the cleanest.

eloycoto:
Hi!

On Flotta we run the worker with a custom user, so we set HOME in the worker.toml ENV variable. 

Regards.

adamruzicka:
> worker.toml ENV variable.

Aren't per-worker configs something that was added post 0.2.0 (which we're kinda stuck on for now)?

eloycoto:
We used master branch.

adamruzicka:
> We're using master branch.

Sadly that is not an option for us at the moment

adamruzicka:
Have we reached an impasse here?

subpop:
I think I'm okay with adding HOME to a worker's environment, and backporting an equivalent feature to 0.2.x, but I'd prefer not to inherit the parent process environment. Are you okay with adding HOME to the worker environment, but setting it to something like '/var/lib/yggdrasil/workers/<worker_name>'?

adamruzicka:
That would help a bit, but also defeat the purpose at the same time. The ultimate goal is to have the same working directory (ie HOME) as when using ssh. Yes, we could then place a symlink from `/var/lib/yggdrasil/workers/<worker name>` to where we need it to be, but that feels needlessly complicated.

subpop:
Oh, so you don't need *a* home directory, you need a *specific* home directory. How tightly bound to the 0.2.x version are you? Are you pulling in whatever is in RHEL AppStream? Are you bound to a 8.x GA compose?

adamruzicka:
Very tightly, we are building and shipping our own builds. At this point in our release cycle, it is not really possible for us to move to a newer version, as that would also involve a change in how workers are registered.

adamruzicka:
We will most likely work around this on our side, but having HOME set properly would still be nice to have.

subpop:
Are you able to update to a patch release of 0.2.x? Like if we backported this, or something like this, to 0.2.x and released it, would you be able to pull the RPM update? Are you using the `rhc` RPM from AppStream or are you building your own RPM?

adamruzicka:
> Are you using the rhc RPM from AppStream or are you building your own RPM?

As I mentioned a bit above, we are not using `rhc` rpm from appstream, we are building our own. The version we build is built from the 0.2.1 tag from this repo. 

> Like if we backported this, or something like this, to 0.2.x and released it, would you be able to pull the RPM update?

We would not be able to pull the rpm update, but we could build our own from the updated sources.

> Are you able to update to a patch release of 0.2.x?

I don't really understand how you do things here. If I'm looking correctly the latest version in this repo is 0.2.98, even though it is not tagged or anything. And somewhere between 0.2.1 and 0.2.98 was a change we can't get over on such a short notice. We might be able to update to 0.2.z for some z, but definitely not for z >= 98. If this could be backported and released in something like 0.2.1.1, then yes, we could update to that.

subpop:
Sorry it isn't clear. The `main` branch is going to eventually become a "0.3.0" release, at which point I will create a release branch (`yggdrasil-0.3`). That branch will continue to receive bug fixes, backports and other patches, but will never receive major or breaking changes. Today, there is a `yggdrsail-0.2` branch that has a similar policy; no breaking changes, but does occasionally get patches, bug fixes and backports. That branch is used to tag and create "0.2.z" releases. The "0.2.98" version you see on `main` is a way of marking the version on the `main` branch as "greater than the 0.2 branch but less than the 0.3 branch".

But if you're building your own RPM and are able to build from the `yggdrasil-0.2` branch, then we do have the ability to deliver backports to you. That's good news.

PR #92, opened 2022-09-09
Simplify http.Client

subpop:
This is part of a series of small refactors that are derived from some experimental work done while exploring other IPC options.

This change lays the groundwork for a network transport refactor that I discovered during the experimentation mentioned above.

Rather than assuming all callers only need the HTTP response body, expose the `http.Response` directly, and let the caller decide how to handle the response. This simplifies the `http.Client` API significantly, and allows for greater flexibility in caller response handling.


subpop:
@eloycoto Could you take a look at the changes I made to the HTTP transport? I tried to preserve the current behavior as best I could figure out, while still refactoring the transport methods to use the updated HTTP client.

eloycoto:
@subpop on PTO currently, I'll have a look early next week

eloycoto (approved):
It took a bit, but it works for us. Thanks!

PR #94, opened 2022-09-22
Configurable worker config directory

subpop:
Add a new config file value and command-line flag: --worker-config-dir.
Setting this flag to a directory path will instruct yggd to look for
worker config files in that directory. The default value is
SYSCONFDIR/yggdrasil/workers, to remain compatible with the previous
implementation.

### DIFF
--- a/cmd/yggd/config.go
+++ b/cmd/yggd/config.go
@@ -10,17 +10,18 @@ import (
 )
 
 const (
-	cliLogLevel      = "log-level"
-	cliCertFile      = "cert-file"
-	cliKeyFile       = "key-file"
-	cliCaRoot        = "ca-root"
-	cliServer        = "server"
-	cliSocketAddr    = "socket-addr"
-	cliClientID      = "client-id"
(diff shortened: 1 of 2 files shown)


cmd/yggd/config.go
```diff
@@ -10,17 +10,18 @@ import (
...
+	cliKeyFile         = "key-file"
+	cliCaRoot          = "ca-root"
+	cliServer          = "server"
+	cliSocketAddr      = "socket-addr"
+	cliClientID        = "client-id"
+	cliPathPrefix      = "path-prefix"
+	cliProtocol        = "protocol"
+	cliDataHost        = "data-host"
+	cliExcludeWorker   = "exclude-worker"
+	cliWorkerConfigDir = "worker-config-dir"
```
strider, line 24: Go fmt will always surprise me. You add one line, and all the structure is reformatted, and part of the commit ... 🤣

strider (approved):
Looks good to me.

PR #97, opened 2022-10-17
fix: Simple fix of Makefile to not require user input

jirihnidek:
* When make is used for building yggdrasil, then gzip is triggered to compress man page, but it required confirmation that old compressed man page would be rewritten. Adding -f option avoid the need of this confirmation.
* Original PR for rhc: https://github.com/RedHatInsights/rhc/pull/19


strider (approved):
LGTM

PR #101, opened 2023-01-30
systemd: Add support for an environment file

cryptomilk:
This will allow to create an environment file like /etc/sysconfig/rhc to set an http proxy or similar. This is needed e.g. for using the internal console staging environment.

### DIFF
--- a/data/systemd/yggd.service.in
+++ b/data/systemd/yggd.service.in
@@ -6,6 +6,7 @@ Requires=network-online.target
 
 [Service]
 Type=simple
+EnvironmentFile=-/etc/sysconfig/@SHORTNAME@
 ExecStart=@SBINDIR@/@SHORTNAME@d
 
 [Install]


cryptomilk:
This way I can easily set a https proxy for connecting internally:

```sh
echo "https_proxy=proxy.internal:3128" > /etc/sysconfig/rhc
```

Ichimonji10:
FWIW, I accomplish this with `systemctl edit rhcd.service`:

```
[Service]
Environment="HTTP_PROXY=http://example.com:3128"
Environment="HTTPS_PROXY=http://example.com:3128"
```

subpop:
I don't know what the trending opinion on `/etc/sysconfig` is either. Is it supposed to be used for new software? Currently, the [recommended way to set up a proxy](https://github.com/RedHatInsights/yggdrasil/wiki/FAQ#how-do-i-connect-to-a-broker-using-an-httphttps-proxy) is to use systemd drop-in files (as mentioned above). While the approach proposed by this PR functionally achieves the same thing, I'd prefer to leave the desired environment file location up to the system administrator and not encourage the use of `/etc/sysconfig` unnecessarily.

PR #103, opened 2023-02-03
Convert dispatcher/worker IPC to D-Bus

subpop:
As proposed in #93, I planned to convert the IPC mechanism `yggd` uses to communicate with workers to D-Bus. When `yggd` starts, it connects to either the system bus or a session bus, depending on the value of the `DBUS_STARTER_BUS_TYPE` environment variable. If that variable has a value of "session", the session bus is used instead of the system bus. This enables multiple instances of `yggd` and workers to run simultaneously on a single host. After connecting to the bus, `yggd` exports itself at the path `/com/redhat/yggdrasil/Dispatcher1`. It implements the `com.redhat.yggdrasil.Dispatcher1` interface. This interface enables workers to send data back across the network transport.`yggd` claims the name `com.redhat.yggdrasil.Dispatcher1`, allowing workers to reach it at a well-known destination. Once all this is done, `yggd` receives messages on its network transport like normal.

Also included in this PR is a new `worker` package. This package encapsulates a lot of the boilerplate code needed to create a worker and connect it to the message bus. The "echo" reference worker implementation has been updated to use this package.

Worker startup is changing with this PR as well. `yggd` is no longer responsible for starting or monitoring worker processes itself. A worker may use whatever startup mechanism it desires; in order to communicate with `yggd`, it must know how to connect to a message bus and claim a well-known name. The `worker` package mentioned above does this automatically, but the exact implementation of how a worker starts up is left to the worker author. It is strongly recommended to install a D-Bus service activation file along with the worker; this will allow the D-Bus broker to start the worker automatically the first time a message is sent to the worker's destination. In addition to a D-Bus service activation file, a worker may opt to include a systemd service. In this case, the broker would defer to systemd to perform the actual worker startup. This permits workers to use any systemd service directives to control their worker execution environment (resource control limits, environment variables, etc.).

*Note*: This is a breaking change. The previous IPC mechanism (gRPC) is removed. Workers *must* be updated to use D-Bus in order to connect to versions of `yggd` going forward.


cmd/yggd/client.go
```diff
@@ -11,41 +11,123 @@ import (
...
+	// start receiving values from the dispatcher and transmit them using the
+	// provided transporter.
+	go func() {
+		for msg := range c.d.Outbound {
+			code, metadata, data, err := c.SendDataMessage(&msg.Data, msg.Data.Metadata)
+			if err != nil {
+				log.Errorf("cannot send data message: %v", err)
+				continue
+			}
+			msg.Resp <- struct {
```
eloycoto, line 53: Shouldn't this struct have a type, it's a bit weird

cmd/yggd/client.go
```diff
@@ -11,41 +11,123 @@ import (
...
+	}
+	return code, metadata, data, nil
 }
 
-func (c *Client) SendEventMessage(msg *yggdrasil.Event) ([]byte, error) {
-	return c.sendMessage(msg, "control")
+func (c *Client) SendEventMessage(msg *yggdrasil.Event) (int, map[string]string, []byte, error) {
+	code, metadata, data, err := c.sendMessage("control", nil, msg)
+	if err != nil {
+		return -1, nil, nil, err
```
eloycoto, line 114: I'm not sure what this  -1 is, could you add a const and return a const that it's easy to read and reduce the cognitive load.

cmd/yggd/client.go
```diff
@@ -11,41 +11,123 @@ import (
...
 	t transport.Transporter
-	d *dispatcher
+	d *work.Dispatcher
 }
 
+// NewClient creates a new Client configured with dispatcher and transporter.
+func NewClient(dispatcher *work.Dispatcher, transporter transport.Transporter) *Client {
+	return &Client{
+		d: dispatcher,
+		t: transporter,
```
eloycoto, line 28: TBH, I would change t to transporter, so reading the rest of the project is a bit easier

internal/sync/mutex_map.go
```diff
@@ -0,0 +1,65 @@
+package sync
+
+import "sync"
+
+// RWMutexMap is a concurrency-safe map, using a sync.RWMutex to lock a backing
+// map when accessing values.
+type RWMutexMap[T any] struct {
```
eloycoto, line 7: I guess that you already checked Go version on RHEL 8 build? Just in case

internal/transport/mqtt.go
```diff
@@ -62,8 +64,10 @@ func NewMQTTTransport(clientID string, broker string, tlsConfig *tls.Config, dat
 		topic = fmt.Sprintf("%v/%v/control/in", yggdrasil.PathPrefix, opts.ClientID())
 		c.Subscribe(topic, 1, func(c mqtt.Client, m mqtt.Message) {
 			go func() {
-				if err := t.ReceiveData(m.Payload(), "control"); err != nil {
-					log.Errorf("cannot receive control message: %v", err)
+				if t.receiveHandler != nil {
```
eloycoto, line 67: I would make if t.receivedHandler == nil {return} so it's a bit flatten

internal/work/dispatcher.go
```diff
@@ -0,0 +1,335 @@
...
+		}),
+	}
+}
+
+// Connect examines the environment for a private D-Bus address and connects to
+// it if found. Otherwise, it connects to the system bus.
+func (d *Dispatcher) Connect() error {
+	var conn *dbus.Conn
+	var err error
+	if os.Getenv("DBUS_SESSION_BUS_ADDRESS") != "" {
```
eloycoto, line 75: Shouldn't this be part of the connect args? Getting ENV variable in the package It's something that I'm not big fan.

internal/sync/mutex_map.go
subpop: I believe as part of this commit, I changed the Go version in go.mod to 1.18. There is no plan to backport this version of yggdrasil to RHEL 8, so RHEL 8 having Go < 1.18 is not an issue.

internal/work/dispatcher.go
subpop: Sort of. `DBUS_SESSION_BUS_ADDRESS` is a well-known environment variable name defined by the [dbus-daemon](https://dbus.freedesktop.org/doc/dbus-daemon.1.html) specification. Because it's set by the broker, I think it's safe to assume to be in the environment of the process. I did discover `DBUS_STARTER_BUS_TYPE` while rereading the man page just now, and I think I can improve this logic to use that variable.

cmd/yggd/client.go
subpop: Good point. I refactored this out into a separate data type.

cmd/yggd/client.go
subpop: Good point. I renamed both `t` and `d` to the full word.

internal/sync/mutex_map.go
eloycoto: Then it's ok :-)

PR #104, opened 2023-02-09
Make canonical-facts opaque data loaded from a file

subpop:
Canonical facts about the host are no longer collected each time a connection-status message is created. Instead, if the administrator wants canonical facts to be part of the connection-status message, they must provide valid canonical facts, structured as JSON, in a file and pass the file to yggd via the canonical-facts-file CLI flag (or corresponding config file option). yggd will unmarshal the content of the file into a map, ensuring it is still valid JSON, but yggd does not inspect the content of the map at all; it merely includes it in the connection-status message.


subpop:
I noticed that `rhc` has a command called 'canonical-facts' that will generate the canonical-facts and print them to JSON. This is the recommended way to generate canonical facts so that yggdrasil can pick them up and use them.

PR #113, opened 2023-04-06
Added a sleep flag to echo worker

ahitacat:
When applied this commit will add a new flag to the echo worker, making it possible to create a slow worker. 

This flag will make the worker to sleep between it receives a message and echoes it.  If the sleep time is not specified it acts as the usual echo worker. 

```sh
[ahitacat@localhost yggdrasil]  go run ./worker/echo --log-level trace --sleep 40s
2023/04/06 14:06:52 addr = echo
2023/04/06 14:06:52 id = ca748f81-2653-4c52-b42b-68793c73d4ca
2023/04/06 14:06:52 responseTo = 
2023/04/06 14:06:52 metadata = map[string]string{}
2023/04/06 14:06:52 data = [104 101 108 108 111]
2023/04/06 14:06:52 emitting event BEGIN
2023/04/06 14:06:52 emitting event WORKING
2023/04/06 14:06:52 sleeping: 40s
2023/04/06 14:07:32 responseCode = 0
2023/04/06 14:07:32 responseMetadata = map[string]string{}
2023/04/06 14:07:32 responseData = []
2023/04/06 14:07:32 emitting event END
```

### DIFF
--- a/worker/echo/main.go
+++ b/worker/echo/main.go
@@ -14,12 +14,20 @@ import (
 	"github.com/redhatinsights/yggdrasil/worker"
 )
 
+var sleepTime time.Duration
+
 // echo opens a new dbus connection and calls the
 // com.redhat.Yggdrasil1.Dispatcher1.Transmit method, returning the metadata and
 // data it received.
 func echo(w *worker.Worker, addr string, id string, responseTo string, metadata map[string]string, data []byte) error {
 	if err := w.EmitEvent(ipc.WorkerEventNameWorking, fmt.Sprintf("echoing %v", data)); err != nil {
 		return fmt.Errorf("cannot call EmitEvent: %w", err)
+	}
+
+	// Sleep time between receiving the message and sending it
+	if sleepTime > 0 {
+		log.Infof("sleeping: %v", sleepTime)
+		time.Sleep(sleepTime)
 	}
 
 	responseCode, responseMetadata, responseData, err := w.Transmit(addr, id, responseTo, metadata, data)
@@ -54,6 +62,7 @@ func main() {
 
 	flag.StringVar(&logLevel, "log-level", "error", "set log level")
 	flag.BoolVar(&remoteContent, "remote-content", false, "connect as a remote content worker")
+	flag.DurationVar(&sleepTime, "sleep", 0, "sleep time in seconds before echoing the response")
 	flag.Parse()
 
 	level, err := log.ParseLevel(logLevel)


worker/echo/main.go
```diff
@@ -54,6 +62,7 @@ func main() {
 
 	flag.StringVar(&logLevel, "log-level", "error", "set log level")
 	flag.BoolVar(&remoteContent, "remote-content", false, "connect as a remote content worker")
+	flag.IntVar(&sleepTime, "sleep", 0, "sleep time in seconds before echoing the response")
```
subpop, line 65: The `flag` package has a [`DurationVar`](https://pkg.go.dev/flag#DurationVar) function that accepts a string that can be parsed to a `time.Duration`. I recommend using this over a plain integer flag. This allows input such as `-sleep 1m` to be valid input.

worker/echo/main.go
```diff
@@ -22,6 +24,12 @@ func echo(w *worker.Worker, addr string, id string, responseTo string, metadata
 		return fmt.Errorf("cannot call EmitEvent: %w", err)
 	}
 
+	// Sleep time between receiving the message and sending it
+	if sleepTime > 0 {
+		log.Tracef("sleeping: %d s", sleepTime)
```
subpop, line 29: This could probably be logged at a lower level, like `Info`. Since it will only ever occur when the user invokes `echo` with a value other than the default, having some console feedback might be acceptable.

subpop (changes requested):

worker/echo/main.go
ahitacat: This is interesting

worker/echo/main.go
```diff
@@ -22,6 +24,12 @@ func echo(w *worker.Worker, addr string, id string, responseTo string, metadata
 		return fmt.Errorf("cannot call EmitEvent: %w", err)
 	}
 
+	// Sleep time between receiving the message and sending it
+	if sleepTime > 0 {
+		log.Infof("sleeping: %v", sleepTime.String())
+		time.Sleep(time.Duration(sleepTime))
```
subpop, line 30: `sleepTime` is already a `time.Duration`. There is no need to wrap it here anymore.

worker/echo/main.go
```diff
@@ -22,6 +24,12 @@ func echo(w *worker.Worker, addr string, id string, responseTo string, metadata
 		return fmt.Errorf("cannot call EmitEvent: %w", err)
 	}
 
+	// Sleep time between receiving the message and sending it
+	if sleepTime > 0 {
+		log.Infof("sleeping: %v", sleepTime.String())
```
subpop, line 29: Using the format parameter `%v` will automatically call a type's `String()` method. There is no need to call it explicitly here.

subpop (changes requested):

subpop (approved):

PR #114, opened 2023-04-13
Added cancel message option to make data message cancellables

ahitacat:
Changes made:

- New parameter to data messages type.
- Cancel parameter is the Message ID of the running message to be cancelled.
- Modify all the methods that used this message type.
- Modify the worker to use the cancel parameter.

When applied this commit it will add a new parameter in the Data Message structure. This parameter will bring the ability to cancel previous messages, thus abort its execution.

The new parameter is a string type, it will contain the `MessageID` of the previous sent message that will be cancelled.
New data messages will have the following format:
```json
{
"type":"data",
"message_id":"5c98d18f-ee97-4089-888b-8bdef3e978f5",
"response_to":"",
"version":1,
"sent":"2023-04-13T12:11:41.802864903+02:00",
"directive":"echo"
,"cancel":"5a78f699-aaee-4103-8f58-9eea7216e2b3",
"metadata":{},
"content":"aGVsbG8="
}
```

The commit modifies the worker adding the ability to cancel an echoing execution. This will only works if it runs in sleep mode.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -182,7 +182,7 @@ for publishing messages and subscribing to topics on an MQTT broker. `mqttcli`
 can be installed with `go install`:
 
 ```
-go install git.sr.ht/~spc/mqttcli/cmd/...
+go install git.sr.ht/~spc/mqttcli/cmd/...@latest
 ```
 
 Or if you're running Fedora 34 or later, it can be installed directly with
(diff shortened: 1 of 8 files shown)


ahitacat:
Close in favor of https://github.com/RedHatInsights/yggdrasil/pull/122

PR #115, opened 2023-05-03
feat: include message id in emitted worker events

DuckBoss:
The purpose of this update is to prepare for the implementation of a 
message journal in yggdrasil by making small changes to workers.
This change adds the worker's message id to the emitted event data in addition 
to the currently existing ipc worker event name and worker message data.

**Current:** `...EmitEvent(worker_event_name, worker_event_message)`
**Proposed:** `...EmitEvent(worker_event_name, worker_event_message, worker_message_id)`

Changes:
- updated dbus interfaces to include new message id field.
- updated "listen" yggctl action to include new message ids.
- updated "WorkerEvent" dbus signals to include new message ids.

### DIFF
--- a/cmd/yggctl/actions.go
+++ b/cmd/yggctl/actions.go
@@ -163,11 +163,11 @@ func listenAction(ctx *cli.Context) error {
 			}
 			name, ok := s.Body[1].(uint32)
 			if !ok {
-				return cli.Exit(fmt.Errorf("cannot cat %T as uint32", s.Body[1]), 1)
+				return cli.Exit(fmt.Errorf("cannot cast %T as uint32", s.Body[1]), 1)
 			}
 			messageID, ok := s.Body[2].(string)
 			if !ok {
-				return cli.Exit(fmt.Errorf("cannot cat %T as string", s.Body[2]), 1)
+				return cli.Exit(fmt.Errorf("cannot cast %T as string", s.Body[2]), 1)
 			}
 			var message string
 			if len(s.Body) > 3 {
--- a/cmd/yggctl/actions.go
+++ b/cmd/yggctl/actions.go
@@ -165,14 +165,18 @@ func listenAction(ctx *cli.Context) error {
 			if !ok {
 				return cli.Exit(fmt.Errorf("cannot cat %T as uint32", s.Body[1]), 1)
 			}
+			messageID, ok := s.Body[2].(string)
+			if !ok {
+				return cli.Exit(fmt.Errorf("cannot cat %T as string", s.Body[2]), 1)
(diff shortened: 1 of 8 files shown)


jirihnidek (approved):
LGTM

ahitacat:
What is the difference between this `message_id` and the id in the `response_to` field?

subpop (changes requested):
This requires a change to the `com.redhat.Yggdrasil1.Worker1` D-Bus interface definition too. That specification defines the fields that can be included in the `Event` signal, and adding a field such as this requires updating the D-Bus specification as well.

worker/worker.go
```diff
@@ -158,9 +158,9 @@ func (w *Worker) Transmit(addr string, id string, responseTo string, metadata ma
 	return
 }
 
-// EmitEvent emits a WorkerEvent, including an optional message.
-func (w *Worker) EmitEvent(event ipc.WorkerEventName, message string) error {
-	args := []interface{}{event}
+// EmitEvent emits a WorkerEvent, worker message id, and an optional message.
+func (w *Worker) EmitEvent(event ipc.WorkerEventName, message string, message_id string) error {
```
subpop, line 162: Could we keep the order of arguments the same as the arguments of the D-Bus method? That reduces cognitive friction when translating between the D-Bus interface description and the implementations.

subpop (changes requested):
This looks great. I just have one small nit-picky thing about function parameter order.

cmd/yggctl/actions.go
```diff
@@ -165,14 +165,18 @@ func listenAction(ctx *cli.Context) error {
 			if !ok {
 				return cli.Exit(fmt.Errorf("cannot cat %T as uint32", s.Body[1]), 1)
 			}
+			messageID, ok := s.Body[2].(string)
+			if !ok {
+				return cli.Exit(fmt.Errorf("cannot cat %T as string", s.Body[2]), 1)
```
subpop, line 170: There is a typo in this log message. Do you mind correcting it?
```suggestion
				return cli.Exit(fmt.Errorf("cannot cast %T as string", s.Body[2]), 1)
```

subpop (changes requested):
I noticed a couple more small changes. One is a typo and one is a variable naming convention. Once those are fixed, I'll merge this. I tested it locally already, so it functions as expected.

cmd/yggctl/actions.go
DuckBoss: Thanks for catching these! I've pushed the changes to fix these.

subpop (approved):

PR #116, opened 2023-05-03
feat: add a message journal

DuckBoss:
This feature adds a message journal that tracks dispatched messages and worker events. 
The `yggctl` command added to enable the retrieval of the message journal can also be filtered
by the message id or the worker name, and worker messages can be truncated by a user specified length.

**Note: This PR is dependent on https://github.com/RedHatInsights/yggdrasil/pull/115**

**Changes:**
 - New `yggctl` comand: `message-journal`
   - Usage: `$ yggctl message-journal <opts>`
   - User specified options:
     - `--truncate-message <number>`: Truncates the worker messages in journal entries that exceed the character count provided.
     - `--message-id <my_message_id>`: Only displays journal entries that have the specified message id.
     - `--worker <my_worker>`: Only displays journal entries that have the specified worker.


jirihnidek (commented):
I have only one request ATM. I will probably have more next week.

cmd/yggctl/main.go
```diff
@@ -250,6 +250,100 @@ func main() {
...
+					Required: false,
+				},
+				&cli.StringFlag{
+					Name:     "message-id",
+					Aliases:  []string{"m"},
+					Usage:    "Only display worker messages and emitted events for the entries with the specified message id.",
+					Required: false,
+				},
+			},
+			Action: func(ctx *cli.Context) error {
```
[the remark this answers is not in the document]
DuckBoss: Thanks for the feedback!
I have updated this PR to pull out the anonymous function inside `main()`.

In this draft [PR#120](https://github.com/RedHatInsights/yggdrasil/pull/120), I have the proposed refactor to move the anonymous action functions in `main()`.
Do you think this is a good way to handle this?

cmd/yggctl/main.go
```diff
@@ -250,6 +250,33 @@ func main() {
...
+		{
+			Name:        "message-journal",
+			Usage:       "Display a list of all worker messages and emitted events",
+			UsageText:   "yggctl message-journal",
+			Description: "The message-journal command sends a dbus call to yggdrasil to retrieve a message journal containing all worker messages and emitted events.",
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:     "truncate-message",
+					Aliases:  []string{"t"},
+					Usage:    "Truncates worker event messages if they exceed the specified character count (10 by default).",
```
subpop, line 262: A default value can automatically be included in the help output by using the `Value` field of a Flag. That sets the default value returned by `c.String("truncate-message")` as well, so you don't need to check the flag value later.

jirihnidek (changes requested):
I have only few small requests ATM. I will test this more tomorrow.

DuckBoss:
### Summary of changes made since last review:
- updated in-memory message journal to use sqlite file to handle runtime and persistent journal entries.
- new filtering flags: `--persistent`, `--from`, `--to`
- adds yggd service flag `--enable-message-journal`
- adds yggd config option `enable-message-journal`
- moved message journal code out of yggd service client into a separate package.
- Added some basic test cases for the updated message journal.

### Changes based on feedback:
**[subpop feedback]**
 - fixed dbus function name mismatch
 - fixed dbus function mismatch with return type
 - dbus spec update to handle journal filtering on the service instead of the yggctl client.
 - added default value 15 to yggctl flag `truncate-message`
 - changed emit event function parameter order.
 
**[jirihnidek feedback]**
 - improved dbus session error handling.

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,216 @@
...
+	if filter.From != "" {
+		queryString += "sent >= $3 "
+		args = append(args, filter.From)
+		if filter.To != "" {
+			queryString += "AND "
+		}
+	}
+	if filter.To != "" {
+		queryString += "sent <= $4 "
+		args = append(args, filter.To)
```
DuckBoss, line 168: Is there a way to create dynamic sql query strings in Go without bringing in an ORM or other external library? It looks messy constructing it with a series of conditional branches...

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil, nil
+	}
+
+	entries := []map[string]string{}
+
+	// Build SQL query to retrieve journal entries.
+	// FIXME: It works but I hate it... is there a better
+	// way to do this without an external library?
+	args := []interface{}{}
+	queryString := "SELECT * FROM "
```
DuckBoss, line 136: Is there a way to create dynamic sql query strings in Go without bringing in an ORM or other external library? It looks messy constructing it with a series of conditional branches...

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,216 @@
...
+	"fmt"
+	"path/filepath"
+	"time"
+
+	_ "github.com/mattn/go-sqlite3"
+	"github.com/redhatinsights/yggdrasil"
+	"github.com/redhatinsights/yggdrasil/ipc"
+)
+
+const messageJournalTemporaryDropString string = `
```
jirihnidek, line 14: I would add some doc string for this variable. Maybe, the name is little bit misleading... I would say that it is not "...String". I would say that it is "...Command"

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,216 @@
...
+	DROP TABLE IF EXISTS runtime;
+`
+const messageJournalTemporaryCreateString string = `
```
jirihnidek, line 17: Same comment here.

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,216 @@
...
+		id INTEGER NOT NULL PRIMARY KEY,
+		message_id VARCHAR(36) NOT NULL,
+		sent DATETIME NOT NULL,
+		worker_name VARCHAR(128) NOT NULL,
+		response_to VARCHAR(36),
+		worker_event INTEGER,
+		worker_message TEXT
+	);
+`
+const messageJournalPersistentCreateString string = `
```
jirihnidek, line 28: Same comment here.

And one more thing. The command is almost the same as for `messageJournalTemporaryCreateString`. Created table contains same rows. The command differs only in `persistent` keyword. I believe this code deserves some refactoring. :-)

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,216 @@
...
+type MessageJournalFilter struct {
+	Persistent     bool
+	TruncateLength int
+	MessageID      string
+	Worker         string
+	From           string
+	To             string
+}
+
+func New(dir string, name string) (*MessageJournal, error) {
```
jirihnidek, line 55: Missing doc-string especially, when this method is exported.

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,216 @@
...
+	);
+`
+
+// MessageJournal is a data structure representing the collection
+// of message journal entries received from worker emitted events and messages.
+type MessageJournal struct {
+	database *sql.DB
+}
+
+type MessageJournalFilter struct {
```
jirihnidek, line 46: Missing doc-string. BTW: My IDE complains that the name starts with package name and it suggests to rename it only to `Filter`.

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil, fmt.Errorf("runtime messsage journal table not created: %w", err)
+	}
+	if _, err = db.Exec(messageJournalPersistentCreateString); err != nil {
+		return nil, fmt.Errorf("persistent messsage journal table not created: %w", err)
+	}
+
+	return &messageJournal, nil
+}
+
+// Adds a new message journal entry to both the temporary runtime table
```
jirihnidek, line 83: The doc string should start with `AddEntry`.

internal/messagejournal/messagejournal.go
jirihnidek: BTW: this variable is used only in function `New()`. Could it be moved to local variable of this function? Or is there any reason, why it is "package" variable?

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,216 @@
...
+// which gets cleared at the start of every session and the persistent table
+// which maintains message entries across multiple sessions.
+func (j *MessageJournal) AddEntry(entry yggdrasil.WorkerMessage) (*yggdrasil.WorkerMessage, error) {
+	// If the message journal is nil, it means the message journal
+	// is not enabled in the config.toml file or provided as a user argument.
+	if j == nil {
+		return nil, fmt.Errorf("message journal not enabled")
+	}
+
+	runtimeAction, err := j.database.Prepare("INSERT INTO runtime(message_id, sent, worker_name, response_to, worker_event, worker_message) values (?,?,?,?,?,?)")
```
jirihnidek, line 93: Again too long lines in this function.

internal/messagejournal/messagejournal.go
jirihnidek: It is good idea to use here something like this.

I'm sorry. I don't have such experience. I think that something like this could be useful, because this function is too long and it could help to refactor this code.

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,216 @@
...
+			"message_id":     messageID,
+			"sent":           sent.String(),
+			"worker_name":    workerName,
+			"response_to":    responseTo,
+			"worker_event":   ipc.WorkerEventName(workerEvent).String(),
+			"worker_message": workerEventMessage,
+		}
+		entries = append(entries, newMessage)
+	}
+	rows.Close()
```
jirihnidek, line 214: The `rows.Close()` returns error and it is not handled.

jirihnidek (changes requested):
Thanks for the PR. It looks promising. :+1: I tried to test it and I crashed the `yggd`. :-) More details in comments. I have several comments and requests.

internal/work/dispatcher.go
```diff
@@ -136,16 +139,56 @@ func (d *Dispatcher) Connect() error {
...
+					}
+					eventMessage, ok := s.Body[2].(string)
+					if !ok {
+						log.Errorf("cannot convert %T to string", s.Body[2])
 						continue
 					}
 					event.Message = eventMessage
+					workerMessageID = eventId
+				case ipc.WorkerEventNameBegin:
+					eventID, ok := s.Body[1].(string)
```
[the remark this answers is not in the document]
subpop: This is correct; the "message" field of an Event is nullable. It's not clearly
annotated as such in the Interface specification, but perhaps it should be.

```
Event:
@name: Name of the event.
@message: Optional message included with the event.
```

internal/work/dispatcher.go
```diff
@@ -136,16 +139,71 @@ func (d *Dispatcher) Connect() error {
 				}
 				event.Name = ipc.WorkerEventName(eventName)
 				event.Worker = strings.TrimPrefix(dest, "com.redhat.Yggdrasil1.Worker1.")
+				workerMessageID := ""
 				switch ipc.WorkerEventName(eventName) {
```
subpop, line 143: I recommend adding a `default` case to this switch that reports an unhandled
event type and continues the loop. You probably don't want execution to continue
when variables like `workerMessageID` are empty strings.

cmd/yggd/main.go
```diff
@@ -236,6 +242,17 @@ func main() {
 			return cli.Exit(fmt.Errorf("unsupported transport protocol: %v", config.DefaultConfig.Protocol), 1)
 		}
 		client := NewClient(dispatcher, transporter)
+
+		enableMessageJournal := config.DefaultConfig.EnableMessageJournal
+		if enableMessageJournal {
+			journal, err := messagejournal.New(constants.ConfigDir, "messagejournal.db")
```
subpop, line 248: Runtime data like this should not be written to the config directory; it belongs
in either `/var` or `/run`. If the configuration option is a filepath, you can
avoid this entirely and rely on the configuration to declare where to write the
message journal database file.

cmd/yggctl/actions.go
```diff
@@ -42,6 +56,81 @@ func generateControlMessageAction(c *cli.Context) error {
 	return nil
 }
 
+func messageJournalAction(ctx *cli.Context) error {
+	// Set the worker message truncate length from the user-provided arguments
+	truncateLength := ctx.Int("truncate-message")
+	if truncateLength <= 0 {
+		return cli.Exit(
+			fmt.Errorf(
```
subpop, line 64: go-staticcheck reports a linting suggestion here:

`error strings should not end with punctuation or newlines (ST1005)go-staticcheck`

cmd/yggctl/actions.go
```diff
@@ -42,6 +56,81 @@ func generateControlMessageAction(c *cli.Context) error {
 	return nil
 }
 
+func messageJournalAction(ctx *cli.Context) error {
+	// Set the worker message truncate length from the user-provided arguments
+	truncateLength := ctx.Int("truncate-message")
```
subpop, line 61: Using a `cli.UintFlag` here will enable you to avoid checking the flag value for negativity.

cmd/yggctl/actions.go
```diff
@@ -42,6 +56,81 @@ func generateControlMessageAction(c *cli.Context) error {
...
+			1,
+		)
+	}
+	// Get the user provided 'worker'/'message-id'/'from'/'to' arguments to filter journal entries if provided.
+	selectedPersistent := ctx.Bool("persistent")
+	selectedWorker := ctx.String("worker")
+	selectedMessageID := ctx.String("message-id")
+	selectedFrom := ctx.String("from")
+	selectedTo := ctx.String("to")
+
```
subpop, line 76: These variables are only accessed once after getting them from the
`cli.Context`. Why not just get them from the context when you need them?

subpop (changes requested):
This is great! I looked over all the changes and have some feedback. Excellent progress though; I like this way this is coming together. What is your reasoning behind two tables in the database? Why track persistent and ephemeral messages separately?

DuckBoss:
> This is great! I looked over all the changes and have some feedback. Excellent progress though; I like this way this is coming together. What is your reasoning behind two tables in the database? Why track persistent and ephemeral messages separately?

I think my reasoning was to maintain simplicity down the line if `persistent` message entries should become an optional feature that can be turned on/off. Apart from that I think it was to just maintain a logical separation, although this wouldn't take much to make into a single table if required.

cmd/yggctl/actions.go
DuckBoss: I thought I made this change... I guess that commit got lost somewhere in the git frenzy. Thanks for pointing this out :+1:

cmd/yggctl/actions.go
DuckBoss: Thanks! Looks like there's another linting tool I didn't have enabled :+1:

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.0 // indirect
 	github.com/gorilla/websocket v1.4.2 // indirect
+	github.com/mattn/go-sqlite3 v1.14.16 // indirect
```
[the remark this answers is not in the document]
DuckBoss: TIL, thanks! :+1:

internal/messagejournal/messagejournal_test.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+	defer os.RemoveAll(tempDir)
+
+	tests := []struct {
+		description string
+		input       [2]string
+	}{
+		{
+			description: "create message journal",
+			input:       [2]string{tempDir, "messagejournal-test.db"},
```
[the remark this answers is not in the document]
DuckBoss: I've run into some weird behavior with in-memory sqlite a long time ago in Python, I'll revisit this.

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,231 @@
...
+		// Convert the entry properties into a string format and append to the list of entries.
+		newMessage := map[string]string{
+			"message_id":     messageID,
+			"sent":           sent.String(),
+			"worker_name":    workerName,
+			"response_to":    responseTo,
+			"worker_event":   ipc.WorkerEventName(workerEvent).String(),
+			"worker_message": workerEventMessage,
+		}
+		entries = append(entries, newMessage)
```
jirihnidek, line 182: It could be very slow, when there will be thousands of entries. Have you tested this case?

internal/messagejournal/migrations/000001_messagejournal.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS journal;
```
jirihnidek, line 1: There is no empty line at the end of the file. Is it expected?

internal/messagejournal/migrations/000001_messagejournal.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS journal (
+    id INTEGER NOT NULL PRIMARY KEY,
+    message_id VARCHAR(36) NOT NULL,
+    sent DATETIME NOT NULL,
+    worker_name VARCHAR(128) NOT NULL,
+    response_to VARCHAR(36),
+    worker_event INTEGER,
+    worker_message TEXT
+);
```
jirihnidek, line 9: Same comment here.

jirihnidek (changes requested):
Thanks for update :+1: I have few comments, questions and requests.

cmd/yggctl/main.go
```diff
@@ -131,6 +131,52 @@ func main() {
...
+					Required: false,
+				},
+				&cli.StringFlag{
+					Name:     "message-id",
+					Aliases:  []string{"m"},
+					Usage:    "Only display worker messages and emitted events for the entries with the specified message id.",
+					Required: false,
+				},
+				&cli.StringFlag{
+					Name:     "from",
```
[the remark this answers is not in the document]
DuckBoss: Yes, that would be better I think. Good idea :)

cmd/yggd/main.go
```diff
@@ -236,6 +242,17 @@ func main() {
 			return cli.Exit(fmt.Errorf("unsupported transport protocol: %v", config.DefaultConfig.Protocol), 1)
 		}
 		client := NewClient(dispatcher, transporter)
+
+		enableMessageJournal := config.DefaultConfig.MessageJournal
+		if enableMessageJournal != "" {
+			journal, err := messagejournal.New(filepath.Join(enableMessageJournal, "messagejournal.db"))
```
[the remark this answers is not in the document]
subpop: I think this whole thing would be simpler if `config.DefaultConfig.MessageJournal` was a path to a file. Then you're not trying to check whether its a directory or not, or blank or not, etc. That also allows for the value `file::memory:` to be given to the option to create an in-memory database.

dbus/com.redhat.Yggdrasil1.xml
```diff
@@ -35,6 +35,39 @@
...
+
+            Returns the set of worker messages dispatched to workers and 
+            events emitted by workers.
+        -->
+        <method name="MessageJournal">
+            <arg type="t" name="truncate_length" direction="in" />
+            <arg type="s" name="message_id" direction="in" />
+            <arg type="s" name="worker" direction="in" />
+            <arg type="s" name="from" direction="in" />
+            <arg type="s" name="to" direction="in" />
```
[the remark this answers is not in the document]
subpop: Yes, 'from' and 'to' should be clarified. I like `from-time` and `to-time`. Another option is the English words "since" and "until". Those have a similar "from time" and "to time" association, respectively.

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,231 @@
...
+		{{if .From}} INTERSECT SELECT * FROM {{.Table}} WHERE sent>='{{.From}}'{{end}}
+		{{if .To}} INTERSECT SELECT * FROM {{.Table}} WHERE sent<='{{.To}}'{{end}}
+		{{if not .Persistent}} INTERSECT SELECT * FROM {{.Table}} WHERE sent>='{{.InitializedAt}}'{{end}}
+		ORDER BY sent`,
+	)
+	if err != nil {
+		return "", fmt.Errorf("cannot parse query template parameters: %w", err)
+	}
+	var compiledQuery bytes.Buffer
+	queryCompileErr := queryTemplateParse.Execute(&compiledQuery,
```
subpop, line 214: You don't need to allocate a new error variable here. You've declared `err`
already, and by the time code reaches this line, you can just assign the return
value of `Execute` to `err`.


```suggestion
	err = queryTemplateParse.Execute(&compiledQuery,
```

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,231 @@
...
+func (j *MessageJournal) GetEntries(filter Filter) ([]map[string]string, error) {
+	entries := []map[string]string{}
+	queryString, err := j.buildDynamicGetEntriesQuery(filter)
+	if err != nil {
+		return nil, fmt.Errorf("cannot build dynamic sql query: %w", err)
+	}
+
+	preparedQuery, err := j.database.Prepare(queryString)
+	if err != nil {
+		return nil, fmt.Errorf("cannot prepare query when retrieving journal entries: %w", err)
```
subpop, line 136: Almost-but-not-quite matching error string text. I like "to retrieve" more than
"when retrieving" personally.

```suggestion
		return nil, fmt.Errorf("cannot prepare query to retrieve journal entries: %w", err)
```

cmd/yggctl/main.go
```diff
@@ -131,6 +131,52 @@ func main() {
...
+					Aliases:  []string{},
+					Usage:    "Only display worker messages and emitted events sent starting from the timestamp provided.",
+					Required: false,
+				},
+				&cli.StringFlag{
+					Name:     "to",
+					Aliases:  []string{},
+					Usage:    "Only display worker messages and emitted events sent up to the timestamp provided.",
+					Required: false,
+				},
```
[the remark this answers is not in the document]
DuckBoss: Sounds good to me :+1:

internal/messagejournal/messagejournal.go
DuckBoss: Yes it could be slow down with thousands of entries, but I was hoping the addition of filters would prevent a case where thousands of events would have to be retrieved at a single time (specifically the `message-id` or `to/sent` filters).

I'm not sure, is there an alternative way to handle this that you think would work well? :thinking:

subpop: I was thinking about this. I wonder if there's a way to defer some of the
aggregation of this function further up the call stack. Maybe instead of
appending each entry to a slice within `GetEntries` and returning it at the end,
`GetEntries` could accept a callback function that gets called on each entry as
its read from the database. That passes the responsibility of aggregating the
results (or not) to the caller. I don't know what advantage that really brings
though, since eventually, this map gets serialized into a D-Bus message and
passed back across the D-Bus broker.

internal/messagejournal/migrations/000001_messagejournal.down.sql
DuckBoss: I haven't seen any functional difference and my editor hasn't complained about it. I can't find any styling guidelines around this case either...maybe just leave it as it is and not add anything extra to migration scripts :shrug:

internal/messagejournal/migrations/000001_messagejournal.up.sql
DuckBoss: See comment above: https://github.com/RedHatInsights/yggdrasil/pull/116#discussion_r1237414637

cmd/yggd/main.go
```diff
@@ -109,6 +110,10 @@ func main() {
 			Usage:  "Wait for `DURATION` before cancelling an HTTP request",
 			Hidden: true,
 		}),
+		altsrc.NewStringFlag(&cli.StringFlag{
+			Name:  config.FlagNameMessageJournal,
+			Usage: "Enables a message journal of emitted worker events and messages at the specified `PATH`.",
```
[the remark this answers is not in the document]
subpop:
```suggestion
			Usage: "Record worker events and messages in the database `FILE`.",
```

cmd/yggd/main.go
```diff
@@ -236,6 +242,17 @@ func main() {
 			return cli.Exit(fmt.Errorf("unsupported transport protocol: %v", config.DefaultConfig.Protocol), 1)
 		}
 		client := NewClient(dispatcher, transporter)
+
+		enableMessageJournal := config.DefaultConfig.MessageJournal
+		if enableMessageJournal != "" {
```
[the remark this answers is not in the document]
subpop: `config.MessageJournal` should be an absolute path to a file.

cmd/yggctl/main.go
```diff
@@ -131,6 +131,52 @@ func main() {
...
+				&cli.StringFlag{
+					Name:     "message-id",
+					Aliases:  []string{"m"},
+					Usage:    "Only display worker messages and emitted events for the entries with the specified message id.",
+					Required: false,
+				},
+				&cli.StringFlag{
+					Name:     "from",
+					Aliases:  []string{},
+					Usage:    "Only display worker messages and emitted events sent starting from the timestamp provided.",
```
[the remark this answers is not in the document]
subpop: The `cli` package does have a `TimestampFlag`. That might help enforce the
proper data type for this flag.

cmd/yggctl/main.go
```diff
@@ -131,6 +131,57 @@ func main() {
...
+		{
+			Name:        "message-journal",
+			Usage:       "Display a list of all worker messages and emitted events",
+			UsageText:   "yggctl message-journal",
+			Description: "The message-journal command sends a dbus call to yggdrasil to retrieve a message journal containing all worker messages and emitted events.",
+			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:     "persistent",
+					Aliases:  []string{"p"},
+					Usage:    "Display worker messages and emitted events gathered across multiple sessions from persistent storage.",
```
subpop, line 143: Nit-pick: Usage strings don't end with periods. They don't need to be complete
sentences. They can be short, descriptive fragments.

cmd/yggctl/main.go
```diff
@@ -131,6 +131,57 @@ func main() {
 			},
 			Action: dispatchAction,
 		},
+		{
+			Name:        "message-journal",
+			Usage:       "Display a list of all worker messages and emitted events",
+			UsageText:   "yggctl message-journal",
+			Description: "The message-journal command sends a dbus call to yggdrasil to retrieve a message journal containing all worker messages and emitted events.",
```
subpop, line 138: D-Bus is an irrelevant implementation detail from the user's perspective. Since
these strings are meant to be user-facing, this description should tell the user
what the operation will/should do, not necessarily how it accomplishes that
operation.

```suggestion
			Description: "The message-journal command retrieves a message list containing all worker messages and emitted events.",
```

cmd/yggctl/main.go
```diff
@@ -131,6 +131,57 @@ func main() {
...
+				&cli.BoolFlag{
+					Name:     "persistent",
+					Aliases:  []string{"p"},
+					Usage:    "Display worker messages and emitted events gathered across multiple sessions from persistent storage.",
+					Required: false,
+				},
+				&cli.UintFlag{
+					Name:     "truncate-message",
+					Aliases:  []string{"t"},
+					Usage:    "Truncates worker event messages if they exceed the specified character count.",
```
subpop, line 149: Try to use the `BACKTICK` notation in Usage text if possible. It helps to
clarify the data types expected in the flag.

cmd/yggctl/actions.go
```diff
@@ -42,6 +56,64 @@ func generateControlMessageAction(c *cli.Context) error {
...
+	}
+	obj := conn.Object("com.redhat.Yggdrasil1", "/com/redhat/Yggdrasil1")
+	if err := obj.Call("com.redhat.Yggdrasil1.MessageJournal", dbus.Flags(0), args...).Store(&journalEntries); err != nil {
+		return cli.Exit(fmt.Errorf("cannot list message journal entries: %v", err), 1)
+	}
+
+	if len(journalEntries) == 0 {
+		fmt.Println("No journal entries found.")
+		return nil
+	}
```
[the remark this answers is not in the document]
subpop: Hm. I think there is value to printing a message if no records are retrieved. If
we don't, we'd either print a single line with the table headers or we'd print
nothing at all. In either case, the user would be left wondering if the command
worked correctly. Reporting an error to the user in the case of zero records
avoids that case.

cmd/yggd/main.go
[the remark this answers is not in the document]
DuckBoss: Hm yes at this point the path might as well directly point to a file.

subpop (changes requested):
Excellent progress! I left another round of comments here and there. Thank you for bearing with me on these rounds of reviews.

dbus/com.redhat.Yggdrasil1.xml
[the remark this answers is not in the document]
DuckBoss: I think I would prefer `since/until` too.. it's less jarring to read in the `--help` output.

cmd/yggctl/main.go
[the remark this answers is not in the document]
DuckBoss: I experimented with the `TimestampFlag` but it was generally more strict about following the provided layout.
So a layout with `YYYY-MM-DD HH:MM:SS` wouldn't allow a user to filter by `YYYY-MM-DD`.

I'll revisit this, maybe it was an implementation error.

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,231 @@
...
+		entry.WorkerName,
+		entry.ResponseTo,
+		entry.WorkerEvent.EventName,
+		entry.WorkerEvent.EventMessage,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("could not insert journal entry into table '%v': %w", messageJournalTableName, err)
+	}
+
+	entryID, err := persistentResult.LastInsertId()
```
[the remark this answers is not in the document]
DuckBoss: This sqlite driver implements `LastInsertId` without issues :+1: 
There's a log trace shortly after that confirms the entry was added along with the insert id and message id.

cmd/yggctl/actions.go
```diff
@@ -42,6 +56,64 @@ func generateControlMessageAction(c *cli.Context) error {
...
+			writer,
+			"%d\t%s\t%s\t%s\t%s\t%v\t%s\n",
+			idx,
+			entry["message_id"],
+			entry["sent"],
+			entry["worker_name"],
+			entry["response_to"],
+			entry["worker_event"],
+			entry["worker_message"],
+		)
```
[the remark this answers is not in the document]
DuckBoss: The worker events being emitted don't contain a `response_to` field, so I instead added an entry from the dispatcher when a message is being dispatched on the worker. In the case of the `echo` worker the `response_to` field isn't being utilized.

Although I think it would be fine if the worker events emitted the "response_to" field as well to prevent this case... any thoughts @subpop ?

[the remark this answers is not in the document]
subpop: Hm. I suppose it's harmless to include the response_to field with events. The worker event already includes the message_id, so a second UUID is reasonable enough.

DuckBoss:
# Summary of Changes... there's a bunch..

## Changes since optional data handling rework:
- test: updated test cases to use new optional event data changes [d55a3f1]
- feat: added 'lastUpdated' time to journal [845b100]
- fix: updated 'since/until' short flags to 's/u' instead of 'ft/tt' for consistency [8b21784]
- **revert: removed `--truncate-message` filter [9972ee9]**
  - I removed this since this option can be greatly improved by changing it to `--truncate-field`, and instead allow the user to truncate whichever optional data field they want instead of only the 'message' field.
  - Above changes can be found in a separate PR: https://github.com/RedHatInsights/yggdrasil/pull/164
- style: formatting updates [7aea134]
- **[Link]** refactor: updated migrations to use sqlite syntax [2fb6938]
- fix: added missing `Required` flag option [d5c393b]
- feat: added `DefaultText` flag param [1791ed7]

## Changes since last review (before optional data handling rework):
The changes made here have all been squashed into [0efddca] and were made before the implementation
of the new method of handling optional event data.
This section might not be useful for reviewers, but provides a more accurate history of changes.
Additionally, the branch with the previous history has been cloned and preserved [HERE](https://github.com/DuckBoss/yggdrasil/tree/jajerome/message-journal-hist)

#### Based on older feedback [commits squashed to 0efddca]:
- **[Link]** refactor: removed unnecessary method return [2b82cd6]
- **[Link/Jiri]** refactor: renamed `to/from-time` CLI flags [c626b3f and c0fa01d]
- **[Link/Jiri]** fix: return error if no entries are found [9134ce5]
- **[Link]** fix: updated wording in message-journal flag to be more user-friendly [037d3c4]
- **[Link]** refactor: updated journal CLI flag to accept a file path [17fd854]


#### Other changes [commits squashed to 0efddca]:
- refactor: simplified error variable names [1920021]
- fix: fixed error message inconsistencies [6f5803b]
- feat: updated optional worker event data to use a map [0c67496]
- feat: added response_to field in emitted events [0c1218c and 09492ad]

subpop (changes requested):
A few small changes here and there; mostly around string content. But great work! I tested the functionality and it works out. I'm going to try and stress test it with some looping message transmission, but I figured I'd get this review back to you to clear up the little bits here and there.

subpop (changes requested):
A few small changes here and there; mostly around string content. But great work! I tested the functionality and it works out. I'm going to try and stress test it with some looping message transmission, but I figured I'd get this review back to you to clear up the little bits here and there.

cmd/yggctl/main.go
```diff
@@ -139,6 +139,53 @@ from stdin.`,
...
+					Aliases:  []string{"u"},
+					Usage:    "Only display worker messages and emitted events sent up to the provided `TIMESTAMP` (YYYY-MM-DD HH:MM:SS)",
+					Required: false,
+				},
+				&cli.StringFlag{
+					Name:        "format",
+					Aliases:     []string{"f"},
+					Usage:       "Print output in `FORMAT` (json, table or text)",
+					Value:       "table",
+					DefaultText: "table",
```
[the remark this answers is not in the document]
DuckBoss: The `DefaultText` is not strictly required, but it specifies what is selected by default in the `--help` output:
```
--format FORMAT, -f FORMAT       Print output in Format (json, table or text) (default: table)
```
I think this is a good idea to add to any parameters that have default values in yggctl.
Doc reference: https://cli.urfave.org/v2/examples/flags/#default-values-for-help-output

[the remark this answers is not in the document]
subpop: Oh, interesting. What happens if `DefaultText` is omitted, but `Value` is set to a string like "table". Doesn't it use that value as the default in the help text? I interpreted `DefaultText` as being an option to override the default value's string formatted value in cases like the example when the default is randomized.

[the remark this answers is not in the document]
DuckBoss: Ah I see.. I checked this again and it looks like that's the case. It looks like it's fine to remove `DefaultText` then.

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,269 @@
...
+type Filter struct {
+	Persistent bool
+	MessageID  string
+	Worker     string
+	Since      string
+	Until      string
+}
+
+type errorJournal struct {
+	err interface{}
```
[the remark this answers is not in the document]
DuckBoss: Being unfamiliar with this, I think I looked to how errors were tested in other parts of the code and used this as a basis:
https://github.com/RedHatInsights/yggdrasil/blob/5130298ed3ac750f3e687cefeb765d869979669d/internal/tags/tags.go#L15-L25

But you're right, using an `error` type and then checking for errors in tests like -
`wantError: &errorJournal{fmt.Errorf("my error message")}` would also work.

internal/messagejournal/messagejournal.go
```diff
@@ -0,0 +1,269 @@
...
+
+// AddEntry adds a new message journal entry to the persistent table
+// in the database.
+func (j *MessageJournal) AddEntry(entry yggdrasil.WorkerMessage) error {
+	const insertEntryTemplate string = `INSERT INTO %s (
+		message_id, sent, worker_name, response_to, worker_event, worker_data)
+		values (?,?,?,?,?,?)`
+
+	insertAction, err := j.database.Prepare(
+		fmt.Sprintf(insertEntryTemplate, messageJournalTableName),
```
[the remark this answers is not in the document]
DuckBoss: Ah good point, this was left over from the previous design.

DuckBoss:
## Changes since last review [10/11/2023]
### yggctl:
- Fix: Removed unnecessary flag option `DefaultText` [d61b1a3]
- Refactor: Updated message-journal flag descriptions [661327a]
### yggd journal:
- Fix:  Fixed an incorrect dbus type `string -> bool` [ed13ffd]
- Refactor: Simplified table name usage, removed table name constants [f30768a]
- Refactor: Updated error struct to use `error` type [36787f6]
- Refactor: Renamed journal 'New' method to 'Open' [4492f4f]
- Refactor: Query builder function moved to be standalone, query string separated into individual lines [3b6dfea] 
    - There were extra newline/tab characters in the query so the query template string has been separated into individual lines. This would cause issues during testing.
- Tests: Added tests for query builder function [e47d50e]

ipc/com.redhat.Yggdrasil1.Worker1.xml
```diff
@@ -88,4 +88,4 @@
             <arg name="out" direction="out" type="s" />
         </method>
     </interface>
-</node>
+</node>
```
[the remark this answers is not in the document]
subpop: This change hasn't been dropped from the PR yet.

DuckBoss:
> I hate to ask, but when was the last time you rebased this onto `main`? It appears to be missing quite a few commits from `main`. I'd be surprised if `go.sum` is the only conflict.

I rebased it somewhat recently  (I think it's been 3 weeks?) so it should be mostly caught up. I'll rebase it again after I make the final adjustments.

cmd/yggd/main.go
```diff
@@ -171,6 +173,29 @@ func setupClient(
 	return client, transporter, nil
 }
 
+// setupMessageJournal tries to set up a message journal database to track
+// worker emitted events at the provided path.
+func setupMessageJournal(client *Client) error {
+	messageJournalPath := config.DefaultConfig.MessageJournal
+	if messageJournalPath != "" {
+		journalFilePath := filepath.Join(messageJournalPath)
```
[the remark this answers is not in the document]
DuckBoss: Honestly, not sure why the `filepath.Join` is still there. It must be another remnant from the previous two-database design. Thanks for catching that.

subpop:
@jirihnidek are your changes addressed? The PR is waiting on your approval, since you made a change request in a previous review.

jirihnidek (changes requested):
There is only one simple to solve conflict with main branch. When conflict is resolved, then it will get my approval. :+1:

jirihnidek (approved):
LGTM, Thanks for all your updates. :+1:

PR #122, opened 2023-05-11
Added new command message: Cancel

ahitacat:
# Proposal

Add a new control message of type command that can be used to cancel a work that is already in progress in a worker. 
`yggd` will receive this command and forward it to the appropriate worker. If this worker implements the cancellation feature it will stop or cancel its work. But it is up to the worker to implement this functionality. 

 This PR also provides an example of how the cancellation message can be implemented in the `echo worker`. It uses a map of synchronized channels to manage the current work, and to send the cancellation message to the `goroutines` that are handle it. 

### This commit adds:

- A new command  message that can be use to cancel work in progress that is done in the workers.Is send as a command to yggdrasil client. Yggdrasil will send this message to the workers Cancel method.

- A new function to the worker, CancelRX. This function will handle the cancel signal in workers side. If it is null the worker doesn't support cancellation of messages.

- echo worker implements this cancel function to handle cancellation when it runs in slow mode.

## How to test it
Follow the [Quickstart](https://github.com/RedHatInsights/yggdrasil/blob/main/CONTRIBUTING.md#quickstart) steps to set the environment. With the following changes:

**Terminal 2**
Run the worker with the loop and sleep parameters, this will allow some time to launch the cancel message.

```
go run ./worker/echo --log-level trace --loop 10 --sleep 10s
```

**Terminal 4**
Run the hello world, that will send a new job to the worker.
```
echo "hello world"| \
go run ./cmd/yggctl generate data-message --directive echo - | \
pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/data/in
```

In the worker it will appear the `id` of that message:
```
2023/10/06 12:21:58 addr = echo
2023/10/06 12:21:58 id = 67c40762-2343-48ad-9bf1-23a9b98bdacb
2023/10/06 12:21:58 responseTo = 
2023/10/06 12:21:58 metadata = map[string]string{}
2023/10/06 12:21:58 data = [104 101 108 108 111 32 119 111 114 108 100 10]
2023/10/06 12:21:58 emitting event BEGIN
```

Use the id **67c40762-2343-48ad-9bf1-23a9b98bdacb** to send the cancellation message:
```
echo '{"command":"cancel", "arguments":{"directive":"echo","messageID":"67c40762-2343-48ad-9bf1-23a9b98bdacb"}}' |\
go run ./cmd/yggctl generate control-message --type command - |\
pub -broker tcp://test.mosquitto.org:1883 -topic yggdrasil/$(hostname)/control/in
```

### DIFF
--- a/cmd/yggd/client.go
+++ b/cmd/yggd/client.go
@@ -318,6 +318,22 @@ func (c *Client) ReceiveControlMessage(msg *yggdrasil.Control) error {
 			if err := c.transporter.Connect(); err != nil {
 				return fmt.Errorf("cannot reconnect to broker: %w", err)
 			}
+		case yggdrasil.CommandNameCancel:
+			log.Info("cancelling message...")
+			// Unmarshall comand arguments
+			// cmd contains the directive and the message id to be canceled.
+			directive, exists := cmd.Arguments["directive"]
+			if !exists {
+				return fmt.Errorf("cancel command does not contain 'directive' argument")
+			}
+			cancelID, exists := cmd.Arguments["messageID"]
+			if !exists {
+				return fmt.Errorf("cancel command does not contain 'messageID' argument")
+			}
+			// Dispatch to appropriate worker.
+			if err := c.dispatcher.CancelMessage(directive, msg.MessageID, cancelID); err != nil {
+				return fmt.Errorf("cannot dispatch cancel message: %w", err)
+			}
 		default:
 			return fmt.Errorf("unknown command: %v", cmd.Command)
 		}
--- a/internal/work/dispatcher.go
+++ b/internal/work/dispatcher.go
@@ -412,3 +412,25 @@ func workerEventFromSignal(s *dbus.Signal) (*ipc.WorkerEvent, error) {
 
 	return &event, nil
 }
+
+// CancelMessage implements the dispatching of a cancel message to the worker.
+func (d *Dispatcher) CancelMessage(directive, message_id, cancel_id string) error {
+	// Send the message through the cancel interface
+	obj := d.conn.Object("com.redhat.Yggdrasil1.Worker1."+directive,
+		dbus.ObjectPath(filepath.Join("/com/redhat/Yggdrasil1/Worker1/", directive)))
+	call := obj.Call("com.redhat.Yggdrasil1.Worker1.Cancel", 0,
+		directive,
+		message_id,
+		cancel_id)
+	if err := call.Store(); err != nil {
+		return fmt.Errorf(
+			"cannot call Cancel method with message %v on worker %v: %v",
+			cancel_id,
+			directive,
+			err,
+		)
+	}
+	log.Debugf("sent cancel message %v to worker %v", cancel_id, directive)
+	d.Dispatchers <- d.FlattenDispatchers()
+	return nil
+}
--- a/worker/echo/main.go
+++ b/worker/echo/main.go
@@ -12,6 +12,7 @@ import (
 
 	"git.sr.ht/~spc/go-log"
 
+	"github.com/redhatinsights/yggdrasil/internal/sync"
 	"github.com/redhatinsights/yggdrasil/ipc"
 	"github.com/redhatinsights/yggdrasil/worker"
 )
@@ -19,10 +20,15 @@ import (
 var sleepTime time.Duration
 var loopIt int
 
-// echo opens a new dbus connection and calls the
-// com.redhat.Yggdrasil1.Dispatcher1.Transmit method, returning the
-// metadata and data. New ID is generated for the message, and
-// response_to is set to the ID of the message we received.
+// syncMapCancelChan is a map of channels that maps message ID and its current
+// work. This allows for the cancellation of a message that has not finished.
+var syncMapCancelChan sync.RWMutexMap[chan struct{}]
+
+// echo handles the echo message and sets the channel that will manage the
+// cancel message. It runs a loop and a sleep according to the loop and
+// sleep parameters, then calls the echo function to transmit the
+// message. If there is a cancellation message during the loop or the sleep
+// time, it will cancel the transmission of the message and finish the work.
 func echo(
 	w *worker.Worker,
 	addr string,
@@ -40,6 +46,9 @@ func echo(
 		return fmt.Errorf("cannot call EmitEvent: %w", err)
 	}
 
+	// Setting the channel to handle cancellation
+	syncMapCancelChan.Set(rcvId, make(chan struct{}))
+
 	// Loop the echoes
 	for i := 0; i < loopIt; i++ {
 		// Sleep time between receiving the message and sending it
@@ -47,32 +56,79 @@ func echo(
 			log.Infof("sleeping: %v", sleepTime)
 			time.Sleep(sleepTime)
 		}
-
-		// Set "response_to" according to the rcvId of the message we received
-		echoResponseTo := rcvId
-		// Create new echoId for the message we are going to send
-		echoId := uuid.New().String()
-
-		responseCode, responseMetadata, responseData, err := w.Transmit(
-			addr,
-			echoId,
-			echoResponseTo,
-			metadata,
-			data,
-		)
-		if err != nil {
-			return fmt.Errorf("cannot call Transmit: %w", err)
+		// Cancel message if it has been sent a cancel message
+		// during sleep time or during the loop
+		cancelChan, _ := syncMapCancelChan.Get(rcvId)
+		select {
+		case <-cancelChan:
+			// if the channel is close delete the channel from map
+			// it will not be longer used.
+			log.Tracef("canceled echo message id: %v", rcvId)
+			log.Tracef("deleting channel from map")
+			syncMapCancelChan.Del(rcvId)
+			return nil
+		default:
+			if err := sendEchoMessage(w, addr, rcvId, responseTo, metadata, data, i); err != nil {
+				return err
+			}
 		}
 
-		// Log the responses received from the Dispatcher, if any.
-		log.Infof("responseCode = %v", responseCode)
-		log.Infof("responseMetadata = %#v", responseMetadata)
-		log.Infof("responseData = %v", responseData)
-		log.Infof("message %v of %v", i+1, loopIt)
+	}
 
-		if err := w.SetFeature("DispatchedAt", time.Now().Format(time.RFC3339)); err != nil {
-			return fmt.Errorf("cannot set feature: %w", err)
-		}
+	log.Tracef("deleting channel from map")
+	syncMapCancelChan.Del(rcvId)
+	return nil
+}
+
+// sendEchoMessage opens a new dbus connection and calls the
+// com.redhat.Yggdrasil1.Dispatcher1.Transmit method, returning the
+// metadata and data. New ID is generated for the message, and
+// response_to is set to the ID of the message we received.
+func sendEchoMessage(
+	w *worker.Worker,
+	addr string,
+	rcvId string,
+	responseTo string,
+	metadata map[string]string,
+	data []byte,
+	count int,
+) error {
+	// Set "response_to" according to the rcvId of the message we received
+	echoResponseTo := rcvId
+	// Create new echoId for the message we are going to send
+	echoId := uuid.New().String()
+
+	responseCode, responseMetadata, responseData, err := w.Transmit(
+		addr,
+		echoId,
+		echoResponseTo,
+		metadata,
+		data,
+	)
+	if err != nil {
+		return fmt.Errorf("cannot call Transmit: %w", err)
+	}
+
+	// Log the responses received from the Dispatcher, if any.
+	log.Infof("responseCode = %v", responseCode)
+	log.Infof("responseMetadata = %#v", responseMetadata)
+	log.Infof("responseData = %v", responseData)
+	log.Infof("message %v of %v", count+1, loopIt)
+
+	if err := w.SetFeature("DispatchedAt", time.Now().Format(time.RFC3339)); err != nil {
+		return fmt.Errorf("cannot set feature: %w", err)
+	}
+	return nil
+}
+
+// cancelEcho receives a cancel message id via com.redhat.Yggdrasil1.Worker1.Cancel method
+// closes the channel associated with that message to cancel its current run
+func cancelEcho(w *worker.Worker, addr string, id string, cancelID string) error {
+	log.Infof("cancelling message with id %v", cancelID)
+	if cancelChan, exists := syncMapCancelChan.Get(cancelID); exists {
+		close(cancelChan)
+	} else {
+		return fmt.Errorf("message with given id: %v does not exist and cannot be canceled", cancelID)
 	}
 
 	return nil
@@ -107,6 +163,7 @@ func main() {
 		"echo",
 		remoteContent,
 		map[string]string{"DispatchedAt": "", "Version": "1"},
+		cancelEcho,
 		echo,
 		events,
 	)
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -16,6 +16,10 @@ import (
 // RxFunc is a function type that gets called each time the worker receives data.
 type RxFunc func(w *Worker, addr string, id string, responseTo string, metadata map[string]string, data []byte) error
 
+// CancelRxFunc is a function type that gets called each time the worker receives
+// a cancel message
+type CancelRxFunc func(w *Worker, addr string, id string, cancelID string) error
+
 // EventHandlerFunc is a function type that gets called each time the worker
 // receives a com.redhat.Yggdrasil1.Dispatcher1.Event signal.
 type EventHandlerFunc func(e ipc.DispatcherEvent)
@@ -26,6 +30,7 @@ type Worker struct {
 	features      map[string]string
 	remoteContent bool
 	rx            RxFunc
+	cancelRx      CancelRxFunc
 	conn          *dbus.Conn
 	objectPath    dbus.ObjectPath
 	busName       string
@@ -37,6 +42,7 @@ func NewWorker(
 	directive string,
 	remoteContent bool,
 	features map[string]string,
+	cancel CancelRxFunc,
 	rx RxFunc,
 	events EventHandlerFunc,
 ) (*Worker, error) {
@@ -49,6 +55,7 @@ func NewWorker(
 		directive:     directive,
 		features:      features,
 		remoteContent: remoteContent,
+		cancelRx:      cancel,
 		rx:            rx,
 		objectPath:    dbus.ObjectPath(path.Join("/com/redhat/Yggdrasil1/Worker1", directive)),
 		busName:       fmt.Sprintf("com.redhat.Yggdrasil1.Worker1.%v", directive),
@@ -101,7 +108,7 @@ func (w *Worker) Connect(quit <-chan os.Signal) error {
 	// Export worker onto the bus, implementing the com.redhat.Yggdrasil1.Worker1
 	// and org.freedesktop.DBus.Introspectable interfaces. The path name the
 	// worker exports includes the directive name.
-	if err := w.conn.ExportMethodTable(map[string]interface{}{"Dispatch": w.dispatch}, w.objectPath, "com.redhat.Yggdrasil1.Worker1"); err != nil {
+	if err := w.conn.ExportMethodTable(map[string]interface{}{"Dispatch": w.dispatch, "Cancel": w.cancel}, w.objectPath, "com.redhat.Yggdrasil1.Worker1"); err != nil {
 		return fmt.Errorf("cannot export com.redhat.Yggdrasil1.Worker1 interface: %w", err)
 	}
 
@@ -194,6 +201,42 @@ func (w *Worker) EmitEvent(
 		dbus.ObjectPath(path.Join("/com/redhat/Yggdrasil1/Worker1", w.directive)),
 		"com.redhat.Yggdrasil1.Worker1.Event",
 		args...)
+}
+
+// cancel implements com.redhat.Yggdrasil1.Worker1.Cancel method by calling the
+// worker's cancelRxFunc in a goroutine.
+func (w *Worker) cancel(addr string, id string, cancelID string) *dbus.Error {
+	// If worker doesn't implement cancellation it does nothing
+	if w.cancelRx == nil {
+		log.Debug("worker does not support cancellation messages")
+		return dbus.NewError(
+			"org.freedesktop.DBus.UnknownInterface",
+			[]interface{}{"cancel method not implemented"},
+		)
+	}
+
+	log.Tracef("cancelling message:")
+	log.Tracef("addr = %v", addr)
+	log.Tracef("id = %v", id)
+	log.Tracef("cancel-id = %v", cancelID)
+
+	// Communicate the worker accepts the message and its starting to
+	// work on it.
+	if err := w.EmitEvent(ipc.WorkerEventNameBegin, id, "", map[string]string{}); err != nil {
+		return dbus.NewError("com.redhat.Yggdrasil1.Worker1.EventError", []interface{}{err.Error()})
+	}
+
+	go func() {
+		if err := w.cancelRx(w, addr, id, cancelID); err != nil {
+			log.Errorf("callback function cancelRx() was terminated: %v", err)
+		}
+		// Communicate to yggd client that the work has finished.
+		if err := w.EmitEvent(ipc.WorkerEventNameEnd, id, "", map[string]string{}); err != nil {
+			log.Errorf("cannot emit event: %v", err)
+		}
+	}()
+
+	return nil
 }
 
 // dispatch implements com.redhat.Yggdrasil1.Worker1.Dispatch by calling the
--- a/ipc/com.redhat.Yggdrasil1.Worker1.xml
+++ b/ipc/com.redhat.Yggdrasil1.Worker1.xml
@@ -27,7 +27,17 @@
             <arg type="a{ss}" name="metadata" direction="in" />
             <arg type="ay" name="data" direction="in" />
         </method>
-
+        <!--
+            Cancel:
+            @directive: worker identifier for which the cancel is destined.
+            @id: unique ID of the message.
+            @cancel_id: unique ID of the message to cancel.
+        -->
+        <method name="Cancel">
+            <arg type="s" name="directive" direction="in" />
+            <arg type="s" name="id" direction="in" />
+            <arg type="s" name="cancel_id" direction="in" />
+        </method>
         <!-- 
             Features:
 
--- a/messages.go
+++ b/messages.go
@@ -43,6 +43,9 @@ const (
 
 	// CommandNameDisconnect instructs a client to permanently disconnect.
 	CommandNameDisconnect CommandName = "disconnect"
+
+	// CommandNameCancel instructs a client to cancel a previous message.
+	CommandNameCancel CommandName = "cancel"
 )
 
 // EventName represents accepted values for the "event" field of an Event


worker/echo/main.go
```diff
@@ -15,19 +15,35 @@ import (
...
 
 	// Sleep time between receiving the message and sending it
 	if sleepTime > 0 {
+		// Setting the channel to handle cancelation
+		channels[id] = make(chan struct{})
+
 		log.Infof("sleeping: %v", sleepTime)
 		time.Sleep(sleepTime)
+
+		// Cancell message if it has been sent a cancel message
```
subpop, line 37:
```suggestion
		// Cancel message if it has been sent a cancel message
```

subpop (changes requested):
This looks great. I'll give this a try to see how it performs too.

internal/work/dispatcher.go
```diff
@@ -319,3 +319,21 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
 
 	return "", fmt.Errorf("cannot get name for sender: %v", sender)
 }
+
+func (d *Dispatcher) CancelMessage(command yggdrasil.Command) error {
+	// Send the message through the cancel interface
+	log.Infof("%v", command)
```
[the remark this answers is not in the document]
ahitacat: There are also similar debug message, I'm changing this to be less verbose.

worker/echo/main.go
```diff
@@ -15,19 +15,35 @@ import (
...
+// in the slow mode.
 func echo(w *worker.Worker, addr string, id string, responseTo string, metadata map[string]string, data []byte) error {
 	if err := w.EmitEvent(ipc.WorkerEventNameWorking, fmt.Sprintf("echoing %v", data)); err != nil {
 		return fmt.Errorf("cannot call EmitEvent: %w", err)
 	}
 
 	// Sleep time between receiving the message and sending it
 	if sleepTime > 0 {
+		// Setting the channel to handle cancelation
+		channels[id] = make(chan struct{})
```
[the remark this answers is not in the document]
ahitacat: Oops, no. According to the documentation maps are not atomic. I'll modify this to work fine with simultaneous goroutines https://go.dev/doc/faq#atomic_maps

worker/worker.go
```diff
@@ -34,6 +38,11 @@ type Worker struct {
 
 // NewWorker creates a new worker.
 func NewWorker(directive string, remoteContent bool, features map[string]string, rx RxFunc, events EventHandlerFunc) (*Worker, error) {
+	w, err := WorkerNew(directive, remoteContent, features, nil, rx, events)
+	return w, err
+}
+
+func WorkerNew(directive string, remoteContent bool, features map[string]string, cancel CancelRxFunc, rx RxFunc, events EventHandlerFunc) (*Worker, error) {
```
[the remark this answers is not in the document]
ahitacat: I was trying to land this feature without the need to change the rest of workers. But probably this doesn't make sense, and it is better just to add that argument to the `WorkerNew` function.

internal/work/dispatcher.go
```diff
@@ -319,3 +319,19 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
...
+func (d *Dispatcher) CancelMessage(command yggdrasil.Command) error {
+	// Send the message through the cancel interface
+	obj := d.conn.Object("com.redhat.Yggdrasil1.Worker1."+command.Content.Arguments["directive"],
+		dbus.ObjectPath(filepath.Join("/com/redhat/Yggdrasil1/Worker1/", command.Content.Arguments["directive"])))
+	call := obj.Call("com.redhat.Yggdrasil1.Worker1.Cancel", 0,
+		command.Content.Arguments["directive"],
+		command.MessageID,
+		command.Content.Arguments["messageID"])
+	if err := call.Store(); err != nil {
+		return fmt.Errorf("cannot call Cancel method on worker: %v", err)
```
jirihnidek, line 332: It would be useful to print here more information about canceling (e.g. worker name: `directive` and `messageID`)

internal/work/dispatcher.go
```diff
@@ -319,3 +319,19 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
 
 	return "", fmt.Errorf("cannot get name for sender: %v", sender)
 }
+
+func (d *Dispatcher) CancelMessage(command yggdrasil.Command) error {
```
jirihnidek, line 323: The method is exported, but it does not have doc-string comment.

worker/echo/main.go
```diff
@@ -47,6 +69,17 @@ func echo(w *worker.Worker, addr string, id string, responseTo string, metadata
 	return nil
 }
 
+func cancelEcho(w *worker.Worker, addr string, id string, cancelID string) error {
```
jirihnidek, line 72: Missing comment with doc-string.

worker/echo/main.go
```diff
@@ -15,19 +16,40 @@ import (
...
 	if err := w.EmitEvent(ipc.WorkerEventNameWorking, fmt.Sprintf("echoing %v", data)); err != nil {
 		return fmt.Errorf("cannot call EmitEvent: %w", err)
 	}
 
 	// Sleep time between receiving the message and sending it
 	if sleepTime > 0 {
-		log.Infof("sleeping: %v", sleepTime)
+		// Setting the channel to handle cancelation
+		chanMutex.Lock()
+		cancelChan[id] = make(chan struct{})
```
jirihnidek, line 35: Isn't too late for creating this channel? Shouldn't we create such channel in main function of this worker before worker is created? It looks like that it will be always necessary to create such channel. Shouldn't be such channel  integrated part of `Worker` structure, and shouldn't we call all of this in `NewWorker()`?

worker/echo/main.go
```diff
@@ -15,19 +16,40 @@ import (
...
+		chanMutex.Lock()
+		select {
+		case <-cancelChan[id]:
+			log.Tracef("canceled echo message id: %v", id)
+			delete(cancelChan, id)
+			return nil
+		default:
+			delete(cancelChan, id)
+		}
+		chanMutex.Unlock()
```
jirihnidek, line 52: Again this looks like as something that will be duplicate in all workers. Shouldn't we create something that will help all developers?

worker/worker.go
```diff
@@ -168,6 +173,36 @@ func (w *Worker) EmitEvent(event ipc.WorkerEventName, message string) error {
 	return w.conn.Emit(dbus.ObjectPath(path.Join("/com/redhat/Yggdrasil1/Worker1", w.directive)), "com.redhat.Yggdrasil1.Worker1.Event", args...)
 }
 
+func (w *Worker) cancel(addr string, id string, cancelID string) *dbus.Error {
```
jirihnidek, line 176: Comment with doc-string is missing.

worker/worker.go
```diff
@@ -168,6 +173,36 @@ func (w *Worker) EmitEvent(event ipc.WorkerEventName, message string) error {
...
+	// If worker doesn't implements the cancelation function it does nothing
+	if w.cancelRX == nil {
+		log.Debug("worker doesn't support cancellation messages")
+		return nil
+	}
+
```
jirihnidek, line 182: I would add here log message introducing following log messages to give them some context. Something like `log.Tracef("cancelling message")`.

jirihnidek (changes requested):
Hi, Thanks for the PR. :+1: I have few, requests, comments and suggestions.

internal/work/dispatcher.go
ahitacat: Do you mean on the error? Or adding a debug message before the `obj.Call`?

worker/echo/main.go
ahitacat: No, we create the map of channels in the main function, which contains the MessageID as the key and the channel as the value. 
The channels of that map should be created when a message arrives at the echo worker and is running with sleep time set. That is a way the worker can control the pending jobs that are running and can be canceled.

This code is the specific implementation of the cancel message for the echo worker. But not all workers will implement such things neither in the same manner.

> Shouldn't be such channel integrated part of the Worker structure, and shouldn't we call all of this in NewWorker()?

Channels are specific to Go. I would try to make the worker the more generally possible to its implementation with libyggd (the C library).

worker/echo/main.go
ahitacat: As previously mentioned, channels are specific to Golang, and this implementation can be different depending on each worker. How to handle the cancellation should be up to each worker.

internal/work/dispatcher.go
```diff
@@ -319,3 +319,19 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
...
+	obj := d.conn.Object("com.redhat.Yggdrasil1.Worker1."+command.Content.Arguments["directive"],
+		dbus.ObjectPath(filepath.Join("/com/redhat/Yggdrasil1/Worker1/", command.Content.Arguments["directive"])))
+	call := obj.Call("com.redhat.Yggdrasil1.Worker1.Cancel", 0,
+		command.Content.Arguments["directive"],
+		command.MessageID,
+		command.Content.Arguments["messageID"])
+	if err := call.Store(); err != nil {
+		return fmt.Errorf("cannot call Cancel method on worker: %v", err)
+	}
+	log.Debugf("sent cancel message %v to worker %v", command.Content.Arguments["messageID"], command.Content.Arguments["directive"])
```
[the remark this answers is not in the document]
ahitacat: I can check messageID, but without directive this method will not be reached. In client.go is when that variable is checked. https://github.com/RedHatInsights/yggdrasil/pull/122/files#diff-c0cb303176b50fa91d326e3a793b98e9ad265858eeec3bf38b4be32b1d924d23R302

internal/work/dispatcher.go
jirihnidek: I mean `fmt.Errorf()`.

internal/work/dispatcher.go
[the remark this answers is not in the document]
jirihnidek: I mean, can we be sure that `command.Content.Arguments["messageID"]` will always exist? Should we be more careful and use something like this?

```go
if messageID, exist := command.Content.Arguments["messageID"]; !exist {
    return fmt.Errorf("cancel command does not contain 'messageID' argument")
}
```

worker/worker.go
jirihnidek: And I added such message (https://github.com/RedHatInsights/yggdrasil/pull/128/files#diff-a497cf86b016d9c95cc8d6a4d99bf1bea36490b188fad3414822222234b654aeR250) ;-)

worker/echo/main.go
jirihnidek: Ah, I can see. The `cancelChan` is created here: https://github.com/RedHatInsights/yggdrasil/pull/122/files#diff-a3f7add2ec318476155f7f939e4242b221a6611a4e50181d2e96fbd13a9ff460R92. Maybe the name `cancelChan` is little bit misleading. What about something like `mapOfCancelChannels`?

internal/work/dispatcher.go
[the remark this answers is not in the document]
ahitacat: Yes I was referring to that with check messageID :smile:

worker/echo/main.go
ahitacat: Make sense if it is not clear. Changing it.

ahitacat:
## Summary of changes made from reviews:
- fix typos in code and comments
- fix logs to follow the conventions on this project and set better debug levels
- `echo worker` use `RWMutexMap` from internal/sync to create a table of channels to handle current pending work. 
    -  Those pending works are related to sent echo message through `yggdrasil`, echo worker running in sleep mode will have the opportunity to cancel that work that has not being finished.
- add doc-string comments

internal/work/dispatcher.go
```diff
@@ -328,3 +328,26 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
 
 	return "", fmt.Errorf("cannot get name for sender: %v", sender)
 }
+
+// CancelMessage implements the dispatch of a cancel method to the worker.
+func (d *Dispatcher) CancelMessage(command yggdrasil.Command, id string) error {
```
jirihnidek, line 333: Following comment seems little bit off.

internal/work/dispatcher.go
```diff
@@ -328,3 +328,26 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
...
+	// Send the message through the cancel interface
+	directive := command.Arguments["directive"]
```
jirihnidek, line 335: Shouldn't we check the existence of `directive` as we do it for `messageID`?

ahitacat: You are right :smile:

jirihnidek (approved):

subpop:
`worker/echo/main.go` has changed; could you resolve those conflicts and we'll see if we can revisit this PR again.

internal/work/dispatcher.go
subpop: This method signature feels odd to me. It strikes me as odd that we expect an
entire `yggdrasil.Command` structure as the first parameter, but then only the
message ID from the encapsulating message as the second parameter. Perhaps the
only parameters that get passed to the dispatcher via this method are the
unpacked values from the command?

```suggestion
func (d *Dispatcher) CancelMessage(directive string, message_id string, cancel_id string) error {
```

This forces all the unmarshalling and parameter validation into the client and
makes it very clear which message ID is to be cancelled.

Another option is to follow the pattern of the `Dispatch` method above, and expect
the entire `yggdrasil.Control` struct in the method:

```suggestion
func (d *Dispatcher) CancelMessage(control yggdrasil.Control) error {
```

This has the downside of unmarshalling the Command struct out of the control
message a second time (the first having occurred in the client when processing
the command name). I'm less inclined to take this approach, because this
introduces the potential to scope-creep the method into a larger "control
message handler" method. It's important to preserve it's responsibility as much
as possible.

So that being said, I think we should push all the message unpacking up to the
client and change the signature of this method as follows:

```suggestion
func (d *Dispatcher) Cancel(directive string, message_id string, cancel_id string) error {
```

worker/worker.go
```diff
@@ -26,14 +29,15 @@ type Worker struct {
 	features      map[string]string
 	remoteContent bool
 	rx            RxFunc
+	cancelRX      CancelRxFunc
```
subpop, line 32: Go naming conventions tend to avoid capitalizations except in very commonly
accepted cases such as "messageID".

```suggestion
	cancelRx      CancelRxFunc
```

worker/worker.go
subpop: Looks like this is settled by the presence of a `log.Tracef` line on 185.

worker/worker.go
```diff
@@ -168,6 +173,39 @@ func (w *Worker) EmitEvent(event ipc.WorkerEventName, messageID string, message
 	return w.conn.Emit(dbus.ObjectPath(path.Join("/com/redhat/Yggdrasil1/Worker1", w.directive)), "com.redhat.Yggdrasil1.Worker1.Event", args...)
 }
 
+// cancel implements com.redhat.Yggdrasil1.Worker1.Cancel method by calling the
+// worker's cancelRXFunc in a goroutine.
+func (w *Worker) cancel(addr string, id string, cancelID string) *dbus.Error {
+	// If worker doesn't implements the cancelation function it does nothing
+	if w.cancelRX == nil {
+		log.Debug("worker does not support cancellation messages")
+		return nil
```
subpop, line 182: I wonder if it's worth responding with a D-Bus error here. This would give yggd
some knowledge about whether the cancellation will work; which it in turn could
pass back across the network transport as an event so that message originators
can be made aware of whether an attempted cancellation will or won't work.

worker/echo/main.go
```diff
@@ -12,16 +12,23 @@ import (
 
 	"git.sr.ht/~spc/go-log"
 
+	"github.com/redhatinsights/yggdrasil/internal/sync"
```
subpop, line 15: This is okay because the echo worker is part of the yggdrasil repository, but I
don't know if Go would permit this import from an external project. I think the
'internal' package has special value to the Go compiler. Might be worth testing.

We could easily solve this by moving the `sync` package to the top of the
`yggdrasil` project, but it seems weird for the `yggdrasil` package to export a
`sync` package.

worker/echo/main.go
```diff
@@ -12,16 +12,23 @@ import (
...
 	"github.com/redhatinsights/yggdrasil/ipc"
 	"github.com/redhatinsights/yggdrasil/worker"
 )
 
 var sleepTime time.Duration
 
+// sycMapCancelChan is a map of channels that maps message ID and its current work
+// this allows to cancel a message that has not finished.
```
subpop, line 23:
```suggestion
// sycMapCancelChan is a map of channels that maps message ID and its current
// work. This allows for the cancellation of a message that has not finished.
```

worker/echo/main.go
```diff
@@ -36,8 +43,23 @@ func echo(
 
 	// Sleep time between receiving the message and sending it
 	if sleepTime > 0 {
-		log.Infof("sleeping: %v", sleepTime)
+		// Setting the channel to handle cancelation
```
subpop, line 46:
```suggestion
		// Setting the channel to handle cancellation
```

worker/echo/main.go
```diff
@@ -36,8 +43,23 @@ func echo(
...
+		cancelChan, _ := sycMapCancelChan.Get(rcvId)
+		select {
+		case <-cancelChan:
+			log.Tracef("canceled echo message id: %v", rcvId)
+			sycMapCancelChan.Del(rcvId)
+			return nil
+		default:
+			sycMapCancelChan.Del(rcvId)
+		}
 	}
```
subpop, line 63: This example feels a bit pedantic to me. Maybe instead of sleeping, the echo
worker could repeatedly echo its message a specific number of times before
"completing" its work. This would allow us to hook a cancel message check into
the work loop, giving a more authentic demonstration of the cancellation.

worker/echo/main.go
```diff
@@ -68,6 +90,17 @@ func echo(
 	return nil
 }
 
+// cancelEcho receives a cancel message id via  com.redhat.Yggdrasil1.Worker1.C.ancel method
+// closes the channel associated to that message to cancel its current run
+func cancelEcho(w *worker.Worker, addr string, id string, cancelID string) error {
+	log.Infof("cancelling message with id %v", cancelID)
+	if cancelChan, exists := sycMapCancelChan.Get(cancelID); exists {
+		close(cancelChan)
```
subpop, line 98: If we've closed the channel because a cancellation was received, the channel
should get deleted from the map too.

subpop (changes requested):
Overall, this is a great approach. We're blazing a new trail here with the introduction of control message data being handled and passed to workers, so I left a few comments about how to maybe improve the API boundary between client and dispatcher. I think we could probably offer a more functional worker example too; considering how the implementation isn't obvious (using channels and sync maps immediately increases the complexity of the example significantly). Having a clear example to use as guidance and documentation is going to be important.

internal/work/dispatcher.go
ahitacat: I wanted to avoid the CancelMessage to unmarshall again the control message. I'm OK to move this logic into the client. It seems a good idea.

cmd/yggd/client.go
```diff
@@ -299,6 +299,13 @@ func (c *Client) ReceiveControlMessage(msg *yggdrasil.Control) error {
 			if err := c.transporter.Connect(); err != nil {
 				return fmt.Errorf("cannot reconnect to broker: %w", err)
 			}
+		case yggdrasil.CommandNameCancel:
+			log.Info("cancelling message...")
+			// Dispatch to apropriated worker.
+			if err := c.dispatcher.CancelMessage(cmd, msg.MessageID); err != nil {
```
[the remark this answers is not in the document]
ahitacat: I needed a way to pass the MessageID of the control message (The new message). The id of the message to be cancel is inside the arguments of the control message.

The Cancel method expects:

`directive` ->  worker name
`id` -> id of the message (same as the newly created)
`messageID` -> to be canceled

This will be more clear with the new arguments that you proposed https://github.com/RedHatInsights/yggdrasil/pull/122/files#r1287363906

worker/echo/main.go
ahitacat: I think this sync package might be useful for worker development, in case other workers want to use this utilty.

worker/worker.go
ahitacat: That makes sense. I think is a good idea that workers reports errors to yggdrasil so it can be aware or some things.

worker/echo/main.go
ahitacat: What do you think to move this pieces of code into the pkg directory?

```
yggdrasil
.
├── build-aux
├── builddir
├── cmd
├── data
├── dbus
├── dist
├── doc
├── internal
├── ipc
├── lib
├── pkg
└── worker
```

That way, we have such an standard way to do the same things. I'm thinking as well in the `http` implementation of the client we have in `yggdrasil`, it can be also be used in `rhc`.

subpop: Yes, we could move `sync` out of `internal` and into the top level of the project. I would prefer to avoid creating a `pkg` package. It is unnecessary and creates an extraneous entry in import paths:

#### With `pkg`

```go
import "github.com/redhatinsights/yggdrasil/pkg/sync"
```

#### Without `pkg`
```go
import "github.com/redhatinsights/yggdrasil/sync"
```

worker/echo/main.go
ahitacat: The channel is deleted from the map when the close message is used, that is in the echo function. The new echo worker functionality may change this.

worker/echo/main.go
```diff
@@ -40,39 +47,83 @@ func echo(
...
 
-		if err := w.SetFeature("DispatchedAt", time.Now().Format(time.RFC3339)); err != nil {
-			return fmt.Errorf("cannot set feature: %w", err)
-		}
+	log.Tracef("Deleting channel from map")
+	sycMapCancelChan.Del(rcvId)
+	return nil
+}
+
+func sendEchoMessage(
```
jirihnidek, line 84: This function does have doc string.

worker/echo/main.go
```diff
@@ -40,39 +47,83 @@ func echo(
...
-			return fmt.Errorf("cannot call Transmit: %w", err)
+		// Cancel message if it has been sent a cancel message
+		// during sleep time or during the loop
+		cancelChan, _ := sycMapCancelChan.Get(rcvId)
+		select {
+		case <-cancelChan:
+			// if the channel is close delete the channel from map
+			// it will not be longer used.
+			log.Tracef("canceled echo message id: %v", rcvId)
+			log.Tracef("Deleting channel from map")
```
jirihnidek, line 68: All log messages should start with lower case.

worker/worker.go
```diff
@@ -196,6 +203,42 @@ func (w *Worker) EmitEvent(
 		args...)
 }
 
+// cancel implements com.redhat.Yggdrasil1.Worker1.Cancel method by calling the
+// worker's cancelRxFunc in a goroutine.
+func (w *Worker) cancel(addr string, id string, cancelID string) *dbus.Error {
+	// If worker doesn't implements the cancelation function it does nothing
```
jirihnidek, line 209: I can see small grammar mistakes ... there should be: "doesn't implement cancellation".

worker/echo/main.go
```diff
@@ -40,39 +47,83 @@ func echo(
...
+	log.Infof("responseData = %v", responseData)
+	log.Infof("message %v of %v", count+1, loopIt)
+
+	if err := w.SetFeature("DispatchedAt", time.Now().Format(time.RFC3339)); err != nil {
+		return fmt.Errorf("cannot set feature: %w", err)
+	}
+	return nil
+}
+
+// cancelEcho receives a cancel message id via  com.redhat.Yggdrasil1.Worker1.C.ancel method
```
jirihnidek, line 121: I can see here some typos: There should be: "cancelEcho receives a cancel message id via com.redhat.Yggdrasil1.Worker1.Cancel method"

worker/echo/main.go
```diff
@@ -40,39 +47,83 @@ func echo(
...
+// closes the channel associated to that message to cancel its current run
+func cancelEcho(w *worker.Worker, addr string, id string, cancelID string) error {
+	log.Infof("cancelling message with id %v", cancelID)
+	if cancelChan, exists := sycMapCancelChan.Get(cancelID); exists {
+		close(cancelChan)
 	}
```
jirihnidek, line 127: When there is no message with given ID, then nothing happens. I would return error here. Something like this:

```go
else {
    return fmt.Errorf("message with given id: %v does not exist and cannot be canceled", cancelID)
)
```

worker/worker.go
```diff
@@ -196,6 +203,42 @@ func (w *Worker) EmitEvent(
...
+
+	// Communicate the worker accepts the message and its starting to
+	// work on it.
+	if err := w.EmitEvent(ipc.WorkerEventNameBegin, id, "", map[string]string{}); err != nil {
+		return dbus.NewError("com.redhat.Yggdrasil1.Worker1.EventError", []interface{}{err.Error()})
+	}
+
+	go func() {
+		if err := w.cancelRx(w, addr, id, cancelID); err != nil {
+			log.Errorf("cannot call cancelRx: %v", err)
```
jirihnidek, line 231: I would re-phrase this error message a little, because it is not 100% correct. I would use something like this:

```go
log.Errorf("callback function cancelRx() was terminated: %v", err)
```

jirihnidek (changes requested):
Thanks for updates :+1: It is better and I think that we are reaching end of PR review process. I have some small suggestions and some requests.

cmd/yggd/client.go
```diff
@@ -318,6 +318,22 @@ func (c *Client) ReceiveControlMessage(msg *yggdrasil.Control) error {
...
+			// cmd contains the directive and the message id to be canceled.
+			directive, exists := cmd.Arguments["directive"]
+			if !exists {
+				return fmt.Errorf("cancel command does not contain 'directive' argument")
+			}
+			cancelID, exists := cmd.Arguments["messageID"]
+			if !exists {
+				return fmt.Errorf("cancel command does not contain 'messageID' argument")
+			}
+			// Dispatch to apropriated worker.
```
subpop, line 333:
```suggestion
			// Dispatch to appropriate worker.
```

worker/worker.go
```diff
@@ -196,6 +203,42 @@ func (w *Worker) EmitEvent(
...
 }
 
+// cancel implements com.redhat.Yggdrasil1.Worker1.Cancel method by calling the
+// worker's cancelRxFunc in a goroutine.
+func (w *Worker) cancel(addr string, id string, cancelID string) *dbus.Error {
+	// If worker doesn't implements the cancelation function it does nothing
+	if w.cancelRx == nil {
+		log.Debug("worker does not support cancellation messages")
+		return dbus.NewError(
+			"org.freedesktop.DBus.UnkownInterface",
```
subpop, line 213: I think this interface name is meant to be `"com.freedesktop.DBus.UnknownInterface"`.

```suggestion
			"org.freedesktop.DBus.UnknownInterface",
```

worker/echo/main.go
```diff
@@ -12,17 +12,23 @@ import (
...
 )
 
 var sleepTime time.Duration
 var loopIt int
 
-// echo opens a new dbus connection and calls the
-// com.redhat.Yggdrasil1.Dispatcher1.Transmit method, returning the
-// metadata and data. New ID is generated for the message, and
-// response_to is set to the ID of the message we received.
+// sycMapCancelChan is a map of channels that maps message ID and its current
```
subpop, line 23:
```suggestion
// syncMapCancelChan is a map of channels that maps message ID and its current
```

subpop (changes requested):
I found one last typo. But the functionality works great!

jirihnidek (approved):
LGTM! Thanks for all updates.

subpop (approved):

PR #124, opened 2023-05-16
feat: refactored yggd main action.

jirihnidek:
* Converted one huge function into several functions.
* The functionality should be almost the same.

### DIFF
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/tls"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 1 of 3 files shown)


subpop (changes requested):
This looks great! I have a pattern question related to `systemdWatchDog`, but otherwise I like the clean up. Thank you!

cmd/yggd/main.go
```diff
@@ -26,6 +27,373 @@ var (
...
+// systemdWatchDog tries to send sd_notify to systemd.
+// More details about sd_notify can be found here:
+// https://www.freedesktop.org/software/systemd/man/sd_notify.html
+func systemdWatchDog() {
+	watchdogDuration, err := daemon.SdWatchdogEnabled(false)
+	if err != nil {
+		log.Errorf("cannot get watchdog duration: %v", err)
+	}
+	if watchdogDuration > 0 {
+		go func() {
```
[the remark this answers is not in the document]
jirihnidek: I also like the pattern that of all these functions are running on goroutines, so I removed the `go func()` here. What do you think?

cmd/yggd/main.go
```diff
@@ -26,6 +27,371 @@ var (
...
+// systemdWatchDog tries to send sd_notify to systemd.
+// More details about sd_notify can be found here:
+// https://www.freedesktop.org/software/systemd/man/sd_notify.html
+func systemdWatchDog() {
+	watchdogDuration, err := daemon.SdWatchdogEnabled(false)
+	if err != nil {
+		log.Errorf("cannot get watchdog duration: %v", err)
+	}
+	if watchdogDuration > 0 {
+		var sdState = daemon.SdNotifyWatchdog
```
subpop, line 275: Why is this assigned to a variable? It doesn't appear like this value will
change between the declaration and the final usage. Wouldn't it be cleaner to
just directly pass `daemon.SdNotifyWatchdog` to `daemon.SdNotify()`?

cmd/yggd/main.go
```diff
@@ -26,6 +27,371 @@ var (
...
+
+// systemdWatchDog tries to send sd_notify to systemd.
+// More details about sd_notify can be found here:
+// https://www.freedesktop.org/software/systemd/man/sd_notify.html
+func systemdWatchDog() {
+	watchdogDuration, err := daemon.SdWatchdogEnabled(false)
+	if err != nil {
+		log.Errorf("cannot get watchdog duration: %v", err)
+	}
+	if watchdogDuration > 0 {
```
subpop, line 274: I feel like I want to see a debug message here.

```suggestion
	if watchdogDuration > 0 {
        log.Debug("starting systemd watchdog notification")
```

subpop (changes requested):
This is great. I think I'd like to see a debug message when the watchdog notification loop starts, but otherwise, I'm ready to merge this.

cmd/yggd/main.go
jirihnidek: Fixed.

cmd/yggd/main.go
jirihnidek: Debug message added.

subpop (approved):

PR #128, opened 2023-05-19
feat: WIP: Added more debug logging to worker

jirihnidek:
* Consider this as WIP, because I did many changes
  in this commit that needs more discussion.
* Added more debug logging messages to worker code
* Wrapped long lines
* Added more comments
* Modified little bit example of echo worker and
  populate id and response_to fields properly
* Added two more events "CONNECTING" and "CONNECTED"
  to show some progress, when worker is starting.
  Without that you cannot see if worker is actually
  "talking" with yggd during start of the worker
  process.
* Fixed code for the case, when worker is started
  in "RemoteContent" mode. Without that it is not
  possible to get response from the HTTP server.
* Tried to fix example worker (echo worker) for the
  case, when it is started with -remote-content CLI
  option. The echo worker tries to use "return_url"
  metadata as a address for w.Transmit().

### DIFF
--- a/internal/work/dispatcher.go
+++ b/internal/work/dispatcher.go
@@ -2,13 +2,6 @@ package work
 
 import (
 	"fmt"
-	"io"
-	"net/url"
-	"os"
-	"path/filepath"
-	"strings"
-	"time"
-
 	"git.sr.ht/~spc/go-log"
 	"github.com/godbus/dbus/v5"
 	"github.com/godbus/dbus/v5/introspect"
@@ -17,6 +10,12 @@ import (
 	internalhttp "github.com/redhatinsights/yggdrasil/internal/http"
 	"github.com/redhatinsights/yggdrasil/internal/sync"
 	"github.com/redhatinsights/yggdrasil/ipc"
+	"io"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
 )
 
 const (
@@ -136,7 +135,9 @@ func (d *Dispatcher) Connect() error {
 				}
 				event.Name = ipc.WorkerEventName(eventName)
 				event.Worker = strings.TrimPrefix(dest, "com.redhat.Yggdrasil1.Worker1.")
+
 				switch ipc.WorkerEventName(eventName) {
+				// Set the event message only if the event is a "working" event.
 				case ipc.WorkerEventNameWorking:
 					eventMessage, ok := s.Body[1].(string)
 					if !ok {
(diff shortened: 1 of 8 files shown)


worker/worker.go
```diff
@@ -24,7 +24,7 @@ type EventHandlerFunc func(e ipc.DispatcherEvent)
 type Worker struct {
 	directive     string
 	features      map[string]string
-	remoteContent bool
+	RemoteContent bool
```
jirihnidek, line 27: Not sure about this change, but It is useful to be able to get this information from the worker. I know that it is possible to get this information using `GetFeature()`, but it makes sense to do it in `yggd`. Doing this in code of worker seems as overkill.

internal/work/dispatcher.go
```diff
@@ -265,36 +275,45 @@ func (d *Dispatcher) Transmit(sender dbus.Sender, addr string, messageID string,
...
 				return TransmitResponseErr, nil, nil, NewDBusError("Transmit", fmt.Sprintf("cannot read HTTP response body: %v", err))
 			}
-			resp.Body.Close()
+			err = resp.Body.Close()
+			if err != nil {
+				return TransmitResponseErr, nil, nil, NewDBusError("Transmit", fmt.Sprintf("cannot close HTTP response body: %v", err))
+			}
+			responseCode = resp.StatusCode
+			// TODO: convert resp.Header to responseMetadata
+			responseData = data
```
jirihnidek, line 284: I was surprised that response from HTTP request was just ignored.

worker/echo/main.go
```diff
@@ -30,7 +36,23 @@ func echo(w *worker.Worker, addr string, id string, responseTo string, metadata
...
-	responseCode, responseMetadata, responseData, err := w.Transmit(addr, id, responseTo, metadata, data)
+	// Set "response_to" according to the id of the message we received
+	echoResponseTo := rcvId
+	// Create new id for the message we are going to send
+	echoId := uuid.New().String()
+	// Create a new echo address, when remote content is enabled
+	echoAddr := ""
+	if w.RemoteContent {
+		if metadata["return_url"] != "" {
+			echoAddr = metadata["return_url"]
```
jirihnidek, line 47: Such approach looks like a convention, but not sure about it too.

internal/work/dispatcher.go
```diff
@@ -2,13 +2,6 @@ package work
 
 import (
 	"fmt"
-	"io"
-	"net/url"
-	"os"
-	"path/filepath"
-	"strings"
-	"time"
-
```
subpop, line 11: This change should not be applied. `goimports` sorts stdlib imports ahead of external packages. They should not be intermixed.

jirihnidek:
This PR is too big. I have created several smaller PRs containing most of changes from this PR. I'm still keeping this for reference.

jirihnidek:
Closing this PR, because all changes here were implemented in smaller PRs.

Issue #136, opened 2023-06-23
Echo worker should set fields in reply messages correctly


jirihnidek:
**Describe the bug**
The example echo worker just send messages as it was received, but it is not correct. Structure of data message is described here: https://github.com/RedHatInsights/yggdrasil/wiki/Protocol#data-messages

**To Reproduce**
1. Start yggd on Terminal 1: `go run ./cmd/yggd --server tcp://localhost:1883 --log-level trace --client-id $(hostname)`
2. Start echo worker on Terminal 2: `go run ./worker/echo -log-level trace`
3. Subscribe to MQTT topics on Terminal 3: `sub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/#`
4. Send some testing MQTT message on Terminal 4: `go run ./cmd/yggctl generate data-message --directive echo "hello" | pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/data/in`

**Current behavior**
You can see following messages on Terminal 3:

Message send from Terminal 4: `2023/06/23 15:55:16 [yggdrasil/thinkpad-p1/data/in]`
```json
{
  "type":"data",
  "message_id":"ebe2e691-1420-4cce-adcc-d1228c6c34d0",
  "response_to":"",
  "version":1,
  "sent":"2023-06-23T15:55:16.136331448+02:00",
  "directive":"echo",
  "metadata":{},
  "content":"aGVsbG8="
}
```
And response from echo worker: `2023/06/23 15:55:16 [yggdrasil/thinkpad-p1/data/out]`
```json
{
  "type":"data",
  "message_id":"ebe2e691-1420-4cce-adcc-d1228c6c34d0",
  "response_to":"",
  "version":1,
  "sent":"2023-06-23T15:55:16.141249454+02:00",
  "directive":"echo",
  "metadata":{},
  "content":"aGVsbG8="
}
```

As you can see the `message_id` is still the same in the response (`ebe2e691-1420-4cce-adcc-d1228c6c34d0`), but `response_to` is not set. 

**Expected behavior**
Response from echo worker should have unique `message_id` and it should set `response_to` to received message.

**Additional Information**
 - OS: Fedora 38
 - Version: main branch 

**Open Questions**
How should echo worker work in remote content mode? Should we use metadata somehow? It looks like that `return_url` should have some special meaning: https://github.com/RedHatInsights/yggdrasil/wiki/Protocol#data-messages, but it is not described there. Documentation contains only example.

[...]

PR #137, opened 2023-06-23
fix: Set properly messageID and responseTo in echo worker

jirihnidek:
* Set fields of transmitted messages properly in echo worker
* This commit should fix issue #136

Fixes: #136


worker/echo/main.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"github.com/google/uuid"
```
subpop, line 6: This import line is in the wrong place. Running `goimports` on the package moves
this line down, outside the list of stdlib imports.

jirihnidek: Fixed.

subpop (approved):

PR #139, opened 2023-06-27
feat: add MQTT connection options

subpop:
Add three options to adjust MQTT client configuration:

    mqtt-connect-retry
    mqtt-connect-retry-interval
    mqtt-auto-reconnect

We are exposing these options as hidden configurable values in the hope that we can identify a cause for some unknown client disconnections.


jirihnidek (approved):
LGTM

PR #140, opened 2023-07-18
feat: reworked optional data sent in emitted events

DuckBoss:
This commit is proposing a change in how optional data is sent in emitted events
and parsed in the yggdrasil dispatcher to simplify the process if additional optional
data needs to be included in emitted worker events in the future.
This update should also not show any changes in behaviour compared to the main branch.

This work primarily came out of attempts to add a `responseTo` optional field in the emitted event
while avoiding parsing complications in the dispatcher. [#141]

### DIFF
--- a/internal/work/dispatcher.go
+++ b/internal/work/dispatcher.go
@@ -396,11 +396,11 @@ func workerEventFromSignal(s *dbus.Signal) (*ipc.WorkerEvent, error) {
 			}
 			event.MessageID = messageID
 		case 2:
-			message, ok := v.(string)
+			data, ok := v.(map[string]string)
 			if !ok {
-				return nil, newStringTypeConversionError(v)
+				return nil, newStringMapTypeConversionError(v)
 			}
-			event.Message = message
+			event.Data = data
 		}
 	}
 
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -176,11 +176,16 @@ func (w *Worker) Transmit(
 	return
 }
 
-// EmitEvent emits a WorkerEvent, worker message id, and an optional message.
-func (w *Worker) EmitEvent(event ipc.WorkerEventName, messageID string, message string) error {
-	args := []interface{}{event, messageID}
-	if message != "" {
-		args = append(args, message)
+// EmitEvent emits a WorkerEvent, worker message id, and key-value pairs of optional data.
+func (w *Worker) EmitEvent(
+	event ipc.WorkerEventName,
+	messageID string,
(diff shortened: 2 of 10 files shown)


worker/worker.go
```diff
@@ -158,11 +158,14 @@ func (w *Worker) Transmit(addr string, id string, responseTo string, metadata ma
 	return
 }
 
-// EmitEvent emits a WorkerEvent, worker message id, and an optional message.
+// EmitEvent emits a WorkerEvent, worker message id, and key-value pairs of optional data.
 func (w *Worker) EmitEvent(event ipc.WorkerEventName, messageID string, message string) error {
```
subpop, line 162: This method signature might as well change to accept a `data map[string]string` parameter instead of `message string`.

internal/work/dispatcher.go
```diff
@@ -144,15 +144,12 @@ func (d *Dispatcher) Connect() error {
...
-						log.Errorf("cannot convert %T to string", s.Body[2])
-						continue
-					}
-					event.Message = eventMessage
+				eventMessageData, ok := s.Body[2].(map[string]string)
+				if !ok {
+					log.Errorf("cannot convert %T to map[string]string", s.Body[2])
+					continue
 				}
+				event.Message = eventMessageData["message"]
```
subpop, line 152: Similarly, the `WorkerEvent` struct should change to have a `data map[string]string` field.

subpop (changes requested):
I like this approach. Since it already breaks API compatibility, we might as well update the struct and EmitEvent method to expect `map[string]string` types instead.

subpop:
Actually, running `golines` across the whole codebase is easy. I'll do that in a PR now.

DuckBoss:
> I appreciate the usage of `golines` to wrap long lines into multiple shorter ones. However, the effective change of this PR is difficult to discern due to the extraneous line wraps. Could you please either keep all the formatting changes into a separate commit, or run `golines` over the entire codebase and open a separate PR updating all the lines in one commit. I think I'd prefer the deliberate PR approach, as it can both keep the changes in a single commit _and_ solve this formatting change for us once. But if you'd prefer a single commit on one PR, I understand.

Yeah I had conflicting thoughts about this too, but I agree running it once over the entire PR would be a better idea. It may be a good idea to update the "contributing" doc to make golines in future PRs mandatory instead of the current optional suggestion and if possible add a golines check to the linting GH action that's currently in use.

subpop (approved):
Looks great!

PR #141, opened 2023-07-18
feat: added "responseTo" field in worker emitted events

DuckBoss:
Currently uses cherry-picked commits from dependent PR: #140 which needs to be removed when the dependent PR gets approved and merged.

### DIFF
--- a/dbus/com.redhat.Yggdrasil1.xml
+++ b/dbus/com.redhat.Yggdrasil1.xml
@@ -40,6 +40,7 @@
             @worker: Name of the worker emitting the event.
             @name: Name of the event.
             @message_id: The id associated with the worker message.
+            @response_to: Unique ID of the message this message is in reply to.
             @data: Key-value pairs of optional data provided with the event.
 
             Emitted by a worker when certain conditions arise, such as beginning
@@ -62,6 +63,7 @@
             <arg type="s" name="worker" />
             <arg type="u" name="name" />
             <arg type="s" name="message_id" />
+            <arg type="s" name="response_to" />
             <arg type="a{ss}" name="data" />
         </signal>
     </interface>
--- a/worker/echo/main.go
+++ b/worker/echo/main.go
@@ -34,6 +34,7 @@ func echo(
 	if err := w.EmitEvent(
 		ipc.WorkerEventNameWorking,
 		rcvId,
+		responseTo,
 		map[string]string{"message": fmt.Sprintf("echoing %v", data)},
 	); err != nil {
 		return fmt.Errorf("cannot call EmitEvent: %w", err)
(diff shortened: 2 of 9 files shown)


dbus/com.redhat.Yggdrasil1.xml
```diff
@@ -62,6 +63,7 @@
             <arg type="s" name="worker" />
             <arg type="u" name="name" />
             <arg type="s" name="message_id" />
+            <arg type="s" name="response_to" />
```
subpop, line 66: We can avoid breaking *this* API by putting the `response_to` field below `message`.

subpop (changes requested):
This should be able to be streamlined if in #140, we replace `message string` with `data map[string]string` everywhere, even in the `WorkerEvent` struct.

worker/echo/main.go
```diff
@@ -31,9 +31,15 @@ func echo(
...
 ) error {
+	// Set "response_to" according to the rcvId of the message we received
+	echoResponseTo := rcvId
+	// Create new echoId for the message we are going to send
+	echoId := uuid.New().String()
+
 	if err := w.EmitEvent(
 		ipc.WorkerEventNameWorking,
-		rcvId,
+		echoId,
```
subpop, line 41: This is interesting. Prior to this change, each event emitted by the worker
would have the message ID of the received message in the first field, resulting
in output that looks like:

```
2023/09/15 12:52:45 echo: 53a8d7d5-e5cb-495c-9d18-2e055d566638: BEGIN: {}
2023/09/15 12:52:45 echo: 53a8d7d5-e5cb-495c-9d18-2e055d566638: WORKING: {"message":"echoing [104 101 108 108 111]"}
2023/09/15 12:52:45 echo: 53a8d7d5-e5cb-495c-9d18-2e055d566638: END: {}
```

Now however, the BEGIN and END events have the originating message ID first:

```
2023/09/15 12:52:45 echo: 53a8d7d5-e5cb-495c-9d18-2e055d566638: BEGIN: : {}
2023/09/15 12:52:45 echo: 80a83707-f982-4e82-82e5-90f06198f70e: WORKING: 53a8d7d5-e5cb-495c-9d18-2e055d566638: {"message":"echoing [104 101 108 108 111]"}
2023/09/15 12:52:45 echo: 53a8d7d5-e5cb-495c-9d18-2e055d566638: END: : {}
```

The two events handled by the `worker` package don't have any awareness to the
"response_to" field, so they don't print anything for that value. The WORKING
event is emitted by the echo worker, and it has created a new UUID for a
response message and puts *that* into the value of WORKING.

Overall, I don't think this is a significant flaw in this PR. But the way its
handled in this echo worker feels conflicted. I think it should continue to emit
the WORKING event, using the message ID of the received message as the first
field to the EmitEvent function. The "responseTo" function parameter has not
changed its meaning; it is still the value of the "response_to" field in the
received message.

subpop (changes requested):

worker/echo/main.go
DuckBoss: Thanks for the feedback!
I've reverted the commit that made this change so now it should continue to emit the message ID of the received message in the EmitEvent function for the **WORKING** event.

subpop (approved):
LGTM

subpop (approved):
LGTM

Issue #142, opened 2023-07-19
yggctl incorrectly connects to session D-Bus socket


subpop:
**Describe the bug**
When logged into a login shell as root, a session D-Bus is started by the seat manager. This results in the `DBUS_SESSION_BUS_ADDRESS` variable exported in the environment. The presence of that variable with a non-empty value causes `yggctl` to connect to the session D-Bus socket instead of the system socket.

**To Reproduce**
1. Log in as root
2. Start `yggd` connecting to the system bus
3. Run `yggctl workers list` as root

**Expected behavior**
The list of workers.

**Actual behavior**
An error occurs: `cannot list workers: The name is not activatable`

jirihnidek:
I wasn't able to reproduce this issue. I tried to reproduce it with `yggdrasil` installed from fedora 38 repository.

```
[root@localhost ~]# rpm -qa | grep yggdrasil
yggdrasil-0.3.1-2.fc38.x86_64
yggdrasil-worker-package-manager-0.2.0-1.fc38.x86_64
```

When I tried to reproduce this issue, then I it does not print any error:

```
[root@localhost ~]# systemctl start yggdrasil.service 

[root@localhost ~]# systemctl status yggdrasil.service 
● yggdrasil.service - yggdrasil system service
     Loaded: loaded (/usr/lib/systemd/system/yggdrasil.service; disabled; preset: disabled)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: active (running) since Thu 2023-07-20 10:21:26 CEST; 1s ago
       Docs: https://github.com/RedHatInsights/yggdrasil
   Main PID: 264502 (yggd)
      Tasks: 11 (limit: 38154)
     Memory: 7.7M
        CPU: 39ms
     CGroup: /system.slice/yggdrasil.service
             └─264502 /usr/bin/yggd

čec 20 10:21:26 thinkpad-p1 systemd[1]: Starting yggdrasil.service - yggdrasil system service...
čec 20 10:21:26 thinkpad-p1 systemd[1]: Started yggdrasil.service - yggdrasil system service.

[root@localhost ~]# yggctl workers list

[root@localhost ~]# systemctl status yggdrasil.service 
● yggdrasil.service - yggdrasil system service
     Loaded: loaded (/usr/lib/systemd/system/yggdrasil.service; disabled; preset: disabled)
    Drop-In: /usr/lib/systemd/system/service.d
[...]

PR #143, opened 2023-07-19
fix(yggctl): prefer system bus when run as root

subpop:
yggctl will attempt to connect to the session bus only if
DBUS_SESSION_BUS_ADDRESS is non-empty and the effective UID of the
process is greater than zero. This results in root invocations of yggctl
preferring the system bus over the session bus.

Fixes: #142

### DIFF
--- a/cmd/yggctl/actions.go
+++ b/cmd/yggctl/actions.go
@@ -43,14 +43,7 @@ func generateControlMessageAction(c *cli.Context) error {
 }
 
 func workersAction(c *cli.Context) error {
-	var conn *dbus.Conn
-	var err error
-
-	if os.Getenv("DBUS_SESSION_BUS_ADDRESS") != "" {
-		conn, err = dbus.ConnectSessionBus()
-	} else {
-		conn, err = dbus.ConnectSystemBus()
-	}
+	conn, err := connectBus()
 	if err != nil {
 		return cli.Exit(fmt.Errorf("cannot connect to bus: %w", err), 1)
 	}
@@ -91,14 +84,7 @@ func workersAction(c *cli.Context) error {
 }
 
 func dispatchAction(c *cli.Context) error {
-	var conn *dbus.Conn
-	var err error
-
-	if os.Getenv("DBUS_SESSION_BUS_ADDRESS") != "" {
-		conn, err = dbus.ConnectSessionBus()
-	} else {
-		conn, err = dbus.ConnectSystemBus()
-	}
+	conn, err := connectBus()
 	if err != nil {
 		return cli.Exit(fmt.Errorf("cannot connect to bus: %w", err), 1)
 	}
@@ -136,14 +122,7 @@ func dispatchAction(c *cli.Context) error {
 }
 
 func listenAction(ctx *cli.Context) error {
-	var conn *dbus.Conn
-	var err error
-
-	if os.Getenv("DBUS_SESSION_BUS_ADDRESS") != "" {
-		conn, err = dbus.ConnectSessionBus()
-	} else {
-		conn, err = dbus.ConnectSystemBus()
-	}
+	conn, err := connectBus()
 	if err != nil {
 		return cli.Exit(fmt.Errorf("cannot connect to bus: %w", err), 1)
 	}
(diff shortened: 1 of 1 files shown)


cmd/yggctl/actions.go
```diff
@@ -209,3 +188,10 @@ func generateMessage(messageType, responseTo, directive, content string, metadat
 		return nil, fmt.Errorf("unsupported message type: %v", messageType)
 	}
 }
+
+func connectBus() (*dbus.Conn, error) {
+	if os.Getenv("DBUS_SESSION_BUS_ADDRESS") != "" && os.Geteuid() > 0 {
```
jirihnidek, line 193: The refactoring is good, but we are loosing chance to print information what bus we tried to connect to. When it is not possible to connect to bus, then we do not important information... Was it session bus or was it system bus? We do not know... I have following proposal:

This block of code is duplicated after all usages of `connectBus()`.:

```go
if err != nil {
	return cli.Exit(fmt.Errorf("cannot connect to bus: %w", err), 1)
}
```

When `error` returned from `connectBus()` is not `nil`, then we can return the error little bit modified. There would be two variants of this error:
1. `fmt.Errorf("cannot connect to system bus: %w", err)`
2. `fmt.Errorf("cannot connect to session %s bus: %w", os.Getenv("DBUS_SESSION_BUS_ADDRESS"), err)`

jirihnidek (changes requested):
I think that refactoring is quite good. I would like to know why the environment variable `DBUS_SESSION_BUS_ADDRESS` was not empty in your case. I think that it is not normal situation. I was not able to reproduce this. See my comments in https://github.com/RedHatInsights/yggdrasil/issues/142 Thus I'm not sure if `os.Geteuid() > 0` is necessary in this case.

cmd/yggctl/actions.go
subpop: Hm. Yea. I can rework the `connectBus` function to return the bus address in the error message.

subpop: I rewrote the `connectBus` function. I hope it is still explicit, but doesn't
lose the bus address information.

cmd/yggctl/actions.go
```diff
@@ -209,3 +188,23 @@ func generateMessage(messageType, responseTo, directive, content string, metadat
...
+	var err error
+	if os.Getenv("DBUS_SESSION_BUS_ADDRESS") != "" && os.Geteuid() > 0 {
+		connect = dbus.ConnectSessionBus
+	} else {
+		connect = dbus.ConnectSystemBus
+	}
+
+	conn, err = connect()
+	if err != nil {
+		if os.Getenv("DBUS_SESSION_BUS_ADDRESS") != "" {
```
jirihnidek, line 204: This line should be changed to:
```go
if os.Getenv("DBUS_SESSION_BUS_ADDRESS") != "" && os.Geteuid() > 0 {
```
otherwise this function will return incorrect error messages for the case you are trying to solve.

jirihnidek: But I have better proposal to avoid checking conditions twice. What about this:
```go
func connectBus() (*dbus.Conn, error) {
	var connect func(...dbus.ConnOption) (*dbus.Conn, error)
	var conn *dbus.Conn
	var err error
	var errMsg string
	if os.Getenv("DBUS_SESSION_BUS_ADDRESS") != "" && os.Geteuid() > 0 {
		connect = dbus.ConnectSessionBus
		errMsg = "cannot connect to session bus (" + os.Getenv("DBUS_SESSION_BUS_ADDRESS") + "): %w"
	} else {
		connect = dbus.ConnectSystemBus
		errMsg = "cannot connect to system bus: %w"
	}

	conn, err = connect()
	if err != nil {
		return nil, fmt.Errorf(errMsg, err)
	}
	return conn, nil
}
```

jirihnidek (changes requested):
I have some request and suggestion.

cmd/yggctl/actions.go
subpop: Nice. Yea. This should work. And I completely forgot my Geteuid check when I rewrote the function. :facepalm:

jirihnidek (approved):
LGTM, Thanks for updates :+1:

PR #145, opened 2023-07-21
Add 'noop' transport

subpop:
Add a noop transport option. Setting the config flag 'protocol' to
"noop" or "" will configure yggd with a no-op transport. All the rest of
the dispatch and routing flows are unaffected, but no data is
transported to or from the client.

### DIFF
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -140,6 +140,16 @@ func setupClient(
 		if err != nil {
 			return nil, nil, cli.Exit(fmt.Errorf("cannot create HTTP transport: %w", err), 1)
 		}
+	case "none":
+		var err error
+		transporter, err = transport.NewNoopTransport()
+		if err != nil {
+			return nil, nil, cli.Exit(fmt.Errorf("cannot create no-op transport: %w", err), 1)
+		}
+		log.Info("no network protocol specified - no data will be sent or received over the network")
+		for _, server := range config.DefaultConfig.Server {
+			log.Warnf("no network protocol specified - ignoring server option '%v'", server)
+		}
 	default:
 		return nil, nil, cli.Exit(
 			fmt.Errorf("unsupported transport protocol: %v", config.DefaultConfig.Protocol),
@@ -436,7 +446,7 @@ func main() {
 		}),
 		altsrc.NewStringFlag(&cli.StringFlag{
 			Name:  config.FlagNameProtocol,
-			Usage: "Transmit data remotely using `PROTOCOL` ('mqtt' or 'http')",
+			Usage: "Transmit data remotely using `PROTOCOL` ('mqtt', 'http' or 'none')",
 			Value: "mqtt",
 		}),
 		altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
--- a/internal/transport/noop.go
+++ b/internal/transport/noop.go
@@ -1,0 +1,33 @@
+package transport
+
+import "crypto/tls"
+
+// Noop is a Transporter that does nothing. This Transport is used to enable a
+// "local only" dispatch mode.
+type Noop struct{}
+
+func NewNoopTransport() (*Noop, error) {
+	return &Noop{}, nil
+}
+
+func (t *Noop) Connect() error {
+	return nil
+}
+
+func (t *Noop) Disconnect(quiesce uint) {}
+
+func (t *Noop) Tx(addr string, metadata map[string]string, data []byte) (responseCode int, responseMetadata map[string]string, responseData []byte, err error) {
+	return
+}
+
+func (t *Noop) SetRxHandler(f RxHandlerFunc) error {
+	return nil
+}
+
+func (t *Noop) ReloadTLSConfig(tlsConfig *tls.Config) error {
+	return nil
+}
+
+func (t *Noop) SetEventHandler(f EventHandlerFunc) error {
+	return nil
+}
--- a/dist/srpm/yggdrasil.spec.in
+++ b/dist/srpm/yggdrasil.spec.in
@@ -102,6 +102,3 @@ export %gomodulesmode
 %{_mandir}/man1/*
 
 %gopkgfiles
-
-%changelog
-%autochangelog


jirihnidek:
What is purpose of such functionality? It seems that `yggd` will be little bit useless in such situation. It looks like that `yggd` would be intended only as dispatcher of messages received over D-Bus, right? Isn't `yggd` overkill in such situation? We can easily send D-Bus methods and signal on localhost without anything like yggdrasil?

I can call `echo` worker, when `yggd` is not running like this:

subpop:
Yes, it does convert `yggd` into little more than a wrapper around D-Bus. I don't have an intended use for this change aside from making my life a little easier as a developer. Being able to turn off the MQTT transport requirements make my ability to develop the other parts of `yggd` a little faster. And who knows? Maybe a customer would find some bizarre use for this. Temporarily stopping transport during maintenance windows, perhaps?

cmd/yggd/main.go
```diff
@@ -140,6 +140,12 @@ func setupClient(
 		if err != nil {
 			return nil, nil, cli.Exit(fmt.Errorf("cannot create HTTP transport: %w", err), 1)
 		}
+	case "", "noop":
```
jirihnidek, line 143: I would use only "noop". I would not use "" as another option for this "protocol". I will explain why. 

Common practice is following. When "" is used for value of some option, then default option is used (default protocol is "mqtt"). Thus suggested behavior would be confusing. The `"noop"` could be used in `/etc/yggdrasil/config.toml` for option `protocol`, right?

BTW: Is the "noop" the best keyword?  I think that it is not so well known word. I tried to find it in some dictionaries and most dictionaries does not contain this word. What about "none" keyword for this purpose? If the "noop" was used, then I would add some description to man page.  The problem is that we generate man page.  Adding additional information to man page would probably require adding custom function for generating man page.

internal/transport/noop.go
```diff
@@ -0,0 +1,33 @@
+package transport
+
+import "crypto/tls"
+
+// Noop is a Transporter that does nothing. This Transport is used to enable a
+// "local only" dispatch mode.
+type Noop struct{}
+
+func NewNoopTransport() (*Noop, error) {
```
jirihnidek, line 9: I would add here at least one info/warning log message saying that we are using "noop" transport "protocol" and nothing will be sent or received.

jirihnidek (changes requested):
I have few comments and suggestions.

jirihnidek:
> Yes, it does convert `yggd` into little more than a wrapper around D-Bus. I don't have an intended use for this change aside from making my life a little easier as a developer. Being able to turn off the MQTT transport requirements make my ability to develop the other parts of `yggd` a little faster. And who knows? Maybe a customer would find some bizarre use for this. Temporarily stopping transport during maintenance windows, perhaps?

OK. It makes sense.

cmd/yggd/main.go
subpop: Hm. Yes. I thought about whether I should include "" or not. You make a good point. I'll drop it from the case condition. I went around and around on a name, including `LoopTransport`, `NullTransport` and a few others. Neither name really fit the functionality of the transport until I settled on the term "no-op". It's an abbreviation for "no operation". I felt this name fit the functionality of the transport quite well, as it doesn't loop anything back; it just takes no operation on the data. I like the term "none" to describe the protocol. Any objections against?

```suggestion
	case "none", "noop":
```

cmd/yggd/main.go
```diff
@@ -436,7 +443,7 @@ func main() {
 		}),
 		altsrc.NewStringFlag(&cli.StringFlag{
 			Name:  config.FlagNameProtocol,
-			Usage: "Transmit data remotely using `PROTOCOL` ('mqtt' or 'http')",
+			Usage: "Transmit data remotely using `PROTOCOL` ('mqtt', 'http' or 'none')",
```
jirihnidek, line 446: The "none" is used in documentation.

cmd/yggd/main.go
```diff
@@ -140,6 +140,13 @@ func setupClient(
 		if err != nil {
 			return nil, nil, cli.Exit(fmt.Errorf("cannot create HTTP transport: %w", err), 1)
 		}
+	case "none", "noop":
+		var err error
+		transporter, err = transport.NewNoopTransport()
+		if err != nil {
+			return nil, nil, cli.Exit(fmt.Errorf("cannot create noop transport: %w", err), 1)
```
jirihnidek, line 147: As you can see. The "noop" key-word is used here.

jirihnidek (changes requested):
I would use "none" or "noop". Not both. It could be confusing to see "noop" and "none" for the same thing.

cmd/yggd/main.go
```diff
@@ -140,6 +140,13 @@ func setupClient(
 		if err != nil {
 			return nil, nil, cli.Exit(fmt.Errorf("cannot create HTTP transport: %w", err), 1)
 		}
+	case "none":
+		var err error
+		transporter, err = transport.NewNoopTransport()
+		if err != nil {
+			return nil, nil, cli.Exit(fmt.Errorf("cannot create no-op transport: %w", err), 1)
+		}
+		log.Info("no network protocol specified; no data will be sent or received over the network")
```
jirihnidek, line 149: When I started `yggd` with `--server` option, then this option was ignored as you can see:

```
[jhnidek@thinkpad-p1 yggdrasil]$ ./builddir/cmd/yggd/yggd --protocol none --server tcp://localhost:1883 --log-level trace --client-id $(hostname)
[yggd] 2023/08/14 10:39:33 /home/jhnidek/github/redhat-insights/yggdrasil/cmd/yggd/main.go:317: starting yggd version 0.3.2
[yggd] 2023/08/14 10:39:33 /home/jhnidek/github/redhat-insights/yggdrasil/cmd/yggd/main.go:149: no network protocol specified; no data will be sent or received over the network
[yggd] 2023/08/14 10:39:33 /home/jhnidek/github/redhat-insights/yggdrasil/internal/work/dispatcher.go:65: connecting to session bus for worker IPC: unix:path=/run/user/1000/bus
[yggd] 2023/08/14 10:39:33 /home/jhnidek/github/redhat-insights/yggdrasil/internal/work/dispatcher.go:92: exported /com/redhat/Yggdrasil1/Dispatcher1 on bus
[yggd] 2023/08/14 10:39:33 /home/jhnidek/github/redhat-insights/yggdrasil/cmd/yggd/client.go:149: connecting to session bus: unix:path=/run/user/1000/bus
[yggd] 2023/08/14 10:39:33 /home/jhnidek/github/redhat-insights/yggdrasil/cmd/yggd/client.go:174: exported /com/redhat/Yggdrasil1 on bus
[yggd] 2023/08/14 10:39:33 /home/jhnidek/github/redhat-insights/yggdrasil/cmd/yggd/main.go:253: no TLS configuration, disabling TLS watcher update
```

I think that it should be good to add there some warning/info message that the `--server` option will be ignored, when `--protoco none` is used. Something like this:

```go
if len(config.DefaultConfig.Server) > 0 {
    log.Infof("server option: %v is ignored for no network protocol", config.DefaultConfig.Server[0]
}
```

jirihnidek (changes requested):
I promise that have final request/comment. :-)

cmd/yggd/main.go
subpop: This starts to get into an area we need to be very cautious. Once flags start becoming mutually exclusive with each other, we begin creating a complicated matrix of flag compatibility. I do not want to maintain a "validateOptions" method that will only grow in complexity as the software matures.

I like your approach though. Rather than stopping the execution of the program with an option validation error, logging a warning that the server value will be ignored can very easily be added in the block of code that sets up the transport.

jirihnidek (approved):
LGTM. :+1: Thanks for updates.

PR #147, opened 2023-07-28
Use systemd PrivateTmp=yes sandboxing

ekohl:
I'll admit I haven't tried this out yet and I have difficulty overseeing all implications. Consider this an RFC with a concrete suggestion on how to implement it.


subpop:
yggd itself doesn't make heavy use of temporary files, but after reading up on PrivateTmp, this does seem like a reasonable thing to enable for the service, with very little downside.

ekohl:
https://www.redhat.com/sysadmin/mastering-systemd may also be a good read on further sandboxing. `systemd-analyze security UNIT` can be used to give recommendations. Note that if you care about EL7 (and I'm pretty sure that's the case) you are limited in what you can use.

PR #148, opened 2023-08-01
fix: handle Event signals gracefully

subpop:
Return errors when handling Event signals emitted by workers and handle
the optional 'message' parameter gracefully.

### DIFF
--- a/internal/work/dispatcher.go
+++ b/internal/work/dispatcher.go
@@ -128,33 +128,14 @@ func (d *Dispatcher) Connect() error {
 					d.Dispatchers <- d.FlattenDispatchers()
 				}
 			case "com.redhat.Yggdrasil1.Worker1.Event":
-				var event ipc.WorkerEvent
-				eventName, ok := s.Body[0].(uint32)
-				if !ok {
-					log.Errorf("cannot convert %T to uint32", s.Body[0])
+				event, err := workerEventFromSignal(s)
+				if err != nil {
+					log.Errorf("cannot unpack signal: %v", err)
 					continue
 				}
-				event.Name = ipc.WorkerEventName(eventName)
 				event.Worker = strings.TrimPrefix(dest, "com.redhat.Yggdrasil1.Worker1.")
 
-				eventMessageID, ok := s.Body[1].(string)
-				if !ok {
-					log.Errorf("cannot convert %T to string", s.Body[1])
-					continue
-				}
-				event.MessageID = eventMessageID
-
-				switch ipc.WorkerEventName(eventName) {
-				case ipc.WorkerEventNameWorking:
-					eventMessage, ok := s.Body[2].(string)
-					if !ok {
-						log.Errorf("cannot convert %T to string", s.Body[2])
-						continue
-					}
-					event.Message = eventMessage
-				}
-
-				d.WorkerEvents <- event
+				d.WorkerEvents <- *event
 			}
 		}
 	}()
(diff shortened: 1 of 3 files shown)


internal/work/errors.go
```diff
@@ -0,0 +1,39 @@
+package work
+
+import "reflect"
+
+var errIncorrectSignal workError = newWorkError("incorrect signal")
```
jirihnidek, line 5: Type `workError` can be omitted.

internal/work/errors.go
```diff
@@ -0,0 +1,39 @@
...
+var errMissingRequiredParameters workError = newWorkError("missing required parameters")
```
jirihnidek, line 6: Same comment here.

internal/work/dispatcher.go
```diff
@@ -340,3 +321,42 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
 
 	return "", fmt.Errorf("cannot get name for sender: %v", sender)
 }
+
+// workerEventFromSignal creates an ipc.WorkerEvent from a DBus signal.
+func workerEventFromSignal(s *dbus.Signal) (*ipc.WorkerEvent, error) {
+	if s.Name != "com.redhat.Yggdrasil1.Worker1.Event" {
+		return nil, errIncorrectSignal
```
jirihnidek, line 328: We are loosing information here, what signal is wrong and what signal is expected. We only return error message: `incorrect signal`. This could be improved.

internal/work/dispatcher.go
```diff
@@ -340,3 +321,42 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
...
+	}
+
+	if len(s.Body) < 2 {
+		return nil, errMissingRequiredParameters
```
jirihnidek, line 332: We could improve error message here and tell user/developer what parameters are required.

internal/work/dispatcher.go
```diff
@@ -340,3 +321,42 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
...
+	}
+
+	event := ipc.WorkerEvent{}
+
+	if len(s.Body) > 0 {
```
jirihnidek, line 337: I think that we can remove these checks for the length of `s.Body`, because we have already checked it

internal/work/dispatcher.go
```diff
@@ -340,3 +321,42 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
...
+		name, ok := s.Body[0].(uint32)
+		if !ok {
+			return nil, newUint32UnpackParameterError(s.Body[0])
+		}
+		event.Name = ipc.WorkerEventName(name)
+	}
+
+	if len(s.Body) > 1 {
```
jirihnidek, line 345: Same comment here.

internal/work/dispatcher.go
```diff
@@ -340,3 +321,42 @@ func (d *Dispatcher) senderName(sender dbus.Sender) (string, error) {
...
+		messageID, ok := s.Body[1].(string)
+		if !ok {
+			return nil, newStringUnpackParameterError(s.Body[1])
+		}
+		event.MessageID = messageID
+	}
+
+	if len(s.Body) > 2 {
```
jirihnidek, line 353: And here.

jirihnidek (changes requested):
Overall, it looks good. :+1: I have few comments and suggestions.

internal/work/dispatcher.go
subpop: I was trying to find a pattern that is recognizable and that clearly indicates which indices of the body slice will be accessed. I know the checks are superfluous, but they serve as sentinels to indicate that certain length slices are going to be used. I might be able to rework this into a switch using fallthrough cases, but I'm not convinced that would more clearly express my intention.

jirihnidek: Hint: you can simply use comments in code ;-).

subpop:
I took a different approach towards unpacking a signal into a WorkerEvent. I dropped the two global variables; they weren't useful without adding variability to them, and after that happens, there is no reason to keep global error variables at all. I renamed the unpack error into a more descriptively named "typeConversionError", since that's more accurately describing what the error carries. I dropped the error checking for signal name and parameter length. Looping over the body slice and attempting to convert the `interface{}` types to their appropriate types will catch any errors that might exist from incorrect signal bodies.

jirihnidek (approved):
LGTM, Thanks for update

PR #149, opened 2023-08-10
feat(yggctl): Read content from files or stdin

subpop:
Update the generate data-message and generate control-message
subcommands to read message content from a file path passed as the first
positional argument on the commandline. Support the convention reading
from stdin if the file path is '-'.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,8 +65,9 @@ Publish a "PING" command to the `yggd` "control/in" topic to test sending
 an MQTT message to `yggd`.
 
 ```
-go run ./cmd/yggctl generate control-message --type command '{"command":"ping"}' | \
-    pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/control/in
+echo '{"command":"ping"}' | \
+go run ./cmd/yggctl generate control-message --type command - | \
+pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/control/in
 ```
 
 You should see the message logged to the output of `sub` in [Terminal
@@ -78,8 +79,9 @@ You should see the message logged to the output of `sub` in [Terminal
 Now publish a data message containing the string "hello" to the `echo` worker.
 
 ```
-go run ./cmd/yggctl generate data-message --directive echo "hello" | \
-    pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/data/in
+echo "hello" | \
+go run ./cmd/yggctl generate data-message --directive echo - | \
+pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/data/in
 ```
 
 Again, you should see the message logged by the `sub` command in [Terminal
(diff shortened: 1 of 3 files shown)


jirihnidek (changes requested):
I personally do not like, when behavior of some command is suddenly changed as it is proposed in this PR. Why didn't you introduce some new flag? Something like `--content-from-file`. I know that `generate` is hidden command, but it is not hidden too much, because it is e.g. mentioned in `CONTRIBUTING.md` BTW: this file has not been changed and instructions there are not valid for this PR, but I hope that I will convince you, and changes in `CONTRIBUTING.md` will not be necessary. ;-)

subpop:
> I personally do not like, when behavior of some command is suddenly changed as it is proposed in this PR. Why didn't you introduce some new flag? Something like `--content-from-file`. I know that `generate` is hidden command, but it is not hidden too much, because it is e.g. mentioned in `CONTRIBUTING.md` BTW: this file has not been changed and instructions there are not valid for this PR, but I hope that I will convince you, and changes in `CONTRIBUTING.md` will not be necessary. ;-)

Generally, I agree with your aversion to changing the behavior of existing commands. However, the benefits this change brings to the program outweigh the cost.

1) The `dispatch` subcommand behaves this way. It expects a file path as the positional command-line argument, and accepts the `-` convention to mean stdin. This PR makes `generate` behave in the same way as `dispatch`. Consistency is good.
2) This change enables the program to be chained into a shell pipeline: `echo hello | yggctl generate data-message -d worker | pub -config mqtt.cfg` (for a developer tool, this is a big benefit).
3) `generate` is an undocumented command that is only useful for developers. I do not mind breaking developers workflows every now and then if it means giving them a better-behaving tool in exchange.

I would much rather make this "API breaking change" now and standardize the file handling conventions, than carry another flag for the rest of the lifespan of this program.

jirihnidek:
> Generally, I agree with your aversion to changing the behavior of existing commands. However, the benefits this change brings to the program outweigh the cost.
> 
>     1. The `dispatch` subcommand behaves this way. It expects a file path as the positional command-line argument, and accepts the `-` convention to mean stdin. This PR makes `generate` behave in the same way as `dispatch`. Consistency is good.
> 
>     2. This change enables the program to be chained into a shell pipeline: `echo hello | yggctl generate data-message -d worker | pub -config mqtt.cfg` (for a developer tool, this is a big benefit).
> 
>     3. `generate` is an undocumented command that is only useful for developers. I do not mind breaking developers workflows every now and then if it means giving them a better-behaving tool in exchange.
> 
> 
> I would much rather make this "API breaking change" now and standardize the file handling conventions, than carry another flag for the rest of the lifespan of this program.

OK, if `dispatch` subcommand uses this approach, then I think that we can go in this way. However, it will be necessary to update file  `CONTRIBUTING.md`, because subcommand `generate` is documented there and developers probably use it.

jirihnidek (approved):
LGTM, Thanks for updates.

PR #156, opened 2023-09-01
ci: added golines lint checking

DuckBoss:
This PR adds golines checking to new PRs and prevents merging to main if it detects that the code has not been formatted for golines.

**Note:**
Not sure if there's a better way to do this, but If there's a way to automatically run golines on every PR during merging instead of requiring commits to comply to golines formatting then that might be a better solution. Of course that could introduce a potential for errors caused by golines to be merged into main so I'm not sure.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -36,6 +36,11 @@ jobs:
         with:
           version: latest
           args: --verbose --timeout=3m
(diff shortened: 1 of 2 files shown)


jirihnidek (approved):
LGTM. I haven't found another way how to run golines as part of CI.

PR #158, opened 2023-09-08
feat: add loop flag to worker

ahitacat:
This commit adds a loop flag to transmit multiple time the echo worker output. By default this flag is 1.

With this flag an user can specified how many messages the echo worker should send.


subpop (approved):

subpop (approved):
LGTM

PR #159, opened 2023-09-14
Set version to 0.4.0

subpop:
The `main` branch currently contains [changes](https://github.com/RedHatInsights/yggdrasil/commit/5f12e93f9edf552e3a1b1a2d77b10185620d78c0) that will break clients expecting the D-Bus protocol that shipped with 0.3.2. While yggdrasil is still "not 1.0", I am trying to maintain as little API breakage as possible. Marking the next version as 0.4.0 hopefully signifies more explicitly that this version is different than previous versions. Clients will need to be updated in order to communicate with yggdrasil 0.4.0.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ["1.18", "1.19"]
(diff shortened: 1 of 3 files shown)


ahitacat (approved):
LGTM, I like the idea to limit API breaks.

PR #162, opened 2023-09-15
Add --mqtt-reconnect-delay flag

subpop:
Add a hidden --mqtt-reconnect-delay flag that will sleep for the given
duration before reconnecting to the broker.


jirihnidek (approved):
LGTM

PR #165, opened 2023-09-26
feat(mqtt): Use Token.WaitTimeout

subpop:
When connecting to a broker and when publishing a message, use the
`Token.WaitTimeout` method instead of `Token.Wait`. `Token.Wait` waits
indefinitely, which can lead to situations when the client never
succeeds in connecting or publishing.

The timeout for each operation can be configured independently by
setting `mqtt-connect-timeout` and `mqtt-publish-timeout`. Both values
default to 30 seconds. The flags are hidden, as they should not commonly
be required to be changed by users.

### DIFF
--- a/internal/transport/mqtt.go
+++ b/internal/transport/mqtt.go
@@ -155,7 +155,15 @@ func (t *MQTT) Connect() error {
 		}
 	}()
 
-	if token := t.client.Connect(); token.Wait() && token.Error() != nil {
+	log.Infof("connecting to broker: %v", config.DefaultConfig.Server)
+	token := t.client.Connect()
+	if !token.WaitTimeout(config.DefaultConfig.MQTTConnectTimeout) {
+		return fmt.Errorf(
+			"cannot connect to broker: connection timeout: %v elapsed",
+			config.DefaultConfig.MQTTConnectTimeout,
+		)
+	}
+	if token.Error() != nil {
 		return fmt.Errorf("cannot connect to broker: %w", token.Error())
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


internal/transport/mqtt.go
```diff
@@ -155,7 +155,12 @@ func (t *MQTT) Connect() error {
 		}
 	}()
 
-	if token := t.client.Connect(); token.Wait() && token.Error() != nil {
+	log.Infof("connecting to broker: %v", config.DefaultConfig.Server)
+	token := t.client.Connect()
+	if !token.WaitTimeout(config.DefaultConfig.MQTTConnectTimeout) {
+		return fmt.Errorf("cannot connect to broker: connection timeout reached")
```
jirihnidek, line 161: I would add the value of timeout to the error message. Something like:

```go
return fmt.Errorf("cannot connect to broker: connection timeout %v reached",
    config.DefaultConfig.MQTTConnectTimeout)
```

internal/transport/mqtt.go
```diff
@@ -190,7 +195,13 @@ func (t *MQTT) Tx(
 	opts := t.client.OptionsReader()
 	topic := fmt.Sprintf("%v/%v/%v/out", config.DefaultConfig.PathPrefix, opts.ClientID(), addr)
 
-	if token := t.client.Publish(topic, 1, false, data); token.Wait() && token.Error() != nil {
+	token := t.client.Publish(topic, 1, false, data)
+	if !token.WaitTimeout(config.DefaultConfig.MQTTPublishTimeout) {
+		return TxResponseErr, nil, nil, fmt.Errorf(
+			"cannot publish message: connection timeout reached",
+		)
```
jirihnidek, line 202: Similar comment here. When you can show some value causing error state, then show it. I would change the code to something like this:

```go
return TxResponseErr, nil, nil, fmt.Errorf(
    "cannot publish message: connection timeout %v reached",
    config.DefaultConfig.MQTTPublishTimeout,
)
```

jirihnidek (changes requested):
Overall LGTM. I have only two small requests.

internal/transport/mqtt.go
subpop: Good point.

jirihnidek (approved):
LGTM, Thanks for update! :+1:

PR #170, opened 2023-11-06
Packaging fixes

ptoscano:
This tweaks the available RPM packaging so it can be used also on EL 8; high-level summary of the changes:
- the Go (S)RPM macros are used only if available, falling back to other macros or setting fallback variables to reduce differences
- generate an upstream tarball compatible with what older RPM versions expect

See the messages of each commit for longer explanations.

### DIFF
--- a/dist/srpm/meson.build
+++ b/dist/srpm/meson.build
@@ -17,7 +17,7 @@ tarball = custom_target('tarball',
     '--gzip',
     '--absolute-names',
     '--file', '@OUTPUT@',
-    '--transform', 's+@0@+@1@-@2@+'.format(meson.project_source_root(), meson.project_name(), commit),
+    '--transform', 's+@0@+@1@-@2@+'.format(meson.project_source_root(), meson.project_name(), meson.project_version()),
     '--exclude', build_root_basename,
     '--exclude', '.git',
     '--exclude', '.vscode',
@@ -25,7 +25,7 @@ tarball = custom_target('tarball',
     '--exclude', '.copr',
     meson.project_source_root()
   ],
-  output: '@0@-@1@.tar.gz'.format(meson.project_name(), commit)
+  output: '@0@-@1@.tar.gz'.format(meson.project_name(), meson.project_version())
 )
 
 specfile = configure_file(
--- a/dist/srpm/yggdrasil.spec.in
+++ b/dist/srpm/yggdrasil.spec.in
@@ -8,6 +8,8 @@
 
 %bcond_without check
 
+%global has_go_rpm_macros (0%{?fedora} || 0%{?rhel} >= 9)
+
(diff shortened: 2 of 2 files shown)


jirihnidek:
Is it necessary to have all of this for the `main` branch? Main branch is used only for Fedora and RHEL8? There is `yggdrasil-0.2` branch for `rhcd` that is used on RHEL8 (https://github.com/RedHatInsights/yggdrasil/tree/yggdrasil-0.2). @

subpop:
> Is it necessary to have all of this for the `main` branch? Main branch is used only for Fedora and RHEL9? There is `yggdrasil-0.2` branch for `rhcd` that is used on RHEL8 (https://github.com/RedHatInsights/yggdrasil/tree/yggdrasil-0.2).

True, though I am packaging `main` code on EL8 in `epel8`, so having `main` build on EL8 is not a bad thing.

subpop (approved):
I'm okay merging this as is, though perhaps @jirihnidek has some additional thoughts, since he was hesitant about merging this onto `main`.

Also, I'm puzzled why the `build_rpm` GitHub Action didn't run.

jirihnidek:
> > Is it necessary to have all of this for the `main` branch? Main branch is used only for Fedora and RHEL9? There is `yggdrasil-0.2` branch for `rhcd` that is used on RHEL8 (https://github.com/RedHatInsights/yggdrasil/tree/yggdrasil-0.2).
> 
> True, though I am [packaging `main` code on EL8](https://src.fedoraproject.org/rpms/yggdrasil/blob/epel8/f/yggdrasil.spec) in `epel8`, so having `main` build on EL8 is not a bad thing.

OK

jirihnidek:
> I'm okay merging this as is, though perhaps @jirihnidek has some additional thoughts, since he was hesitant about merging this onto `main`.
> 
> Also, I'm puzzled why the `build_rpm` GitHub Action didn't run.

Couldn't be the reason that the PR was created from Pino's fork? I can see that this action haven't been triggered for any open PR (all of them are from forks).

jirihnidek:
@subpop @ptoscano I thought that building of RPM was tested during CI. Does anybody tested building RPM manually and how was it tested? Can we trigger `build-srpm` and `build-rpm` manually?

ptoscano:
> I thought that building of RPM was tested during CI.

According to the current actions, the RPM build jobs are run only on push:
https://github.com/RedHatInsights/yggdrasil/blob/522e6ae5db5f12d579bbf409bb388d7de5979283/.github/workflows/package.yml#L3-L4

> Can we trigger build-srpm and build-rpm manually?

Considering that the action (see link above) does not have `workflow_dispatch:`, it cannot be run manually.

Also, the plan is to use Packit here as well, hopefully happening soon.

PR #184, opened 2023-11-09
fix: add http env to workers execution

ahitacat:
yggdrasil adds the proxy as an env variable when starts the worker.

Resolves: CCT-112


HACKING.md
```diff
@@ -59,6 +59,10 @@ export CONSUMER_ID=$(openssl x509 -in cert.pem -subject -nocert | cut -f3 -d" ")
 mosquitto_pub --host 127.0.0.1 --port 1883 --topic "yggdrasil/${CONSUMER_ID}/control/in" --message "{\"type\":\"command\",\"message_id\":\"$(uuidgen | tr -d '\n')\",\"version\":1,\"sent\":\"$(date --iso-8601=seconds --utc | tr -d '\n')\",\"content\":{\"command\":\"ping\"}}"
 ```
 
+echo '{"command":"cancel", "arguments":{"directive":"echo","messageID":"dbc8c1a7-8552-4b79-b36d-723aa68c25dc"}}' | go run ./cmd/yggctl generate control-message --type command - | pub -broker tcp://test.mosquitto.org:1883 -topic yggdrasil/$(hostname)/control/in
```
jirihnidek, line 62: Can you please add some description of this new commands? This change should be in another commit, because it looks that it is not related to the fix.

cmd/yggd/main.go
```diff
@@ -377,6 +377,9 @@ func main() {
 		env := []string{
 			"YGG_SOCKET_ADDR=unix:" + c.String("socket-addr"),
 			"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
```
jirihnidek, line 379: @subpop is it good idea to have `PATH` hard-coded like this? Wouldn't be better to get current `PATH` using `os.Getenv("PATH")`?

jirihnidek (changes requested):
Overall it looks OK. I have only small requests.

HACKING.md
ahitacat: Oh this is a wrong line, I'm not sure how it got there. Deleting it.

cmd/yggd/main.go
ahitacat: Just to point here, when using os.Getenv("PATH") workers has the following path: `PATH=/sbin:/bin:/usr/sbin:/usr/bin` 
So maybe changing that could cause some issues on current workers that they expect to use local paths.

cmd/yggd/main.go
```diff
@@ -377,6 +377,9 @@ func main() {
...
+			"HTTP_PROXY=" + os.Getenv("HTTP_PROXY"),
```
subpop, line 380: The history of these environment variable names is messy. `HTTP_PROXY` might be supported by some implementations, but the "curl standard" is that `http_proxy` is lower-case, so we should include the lower and upper-case values.

https://everything.curl.dev/usingcurl/proxies/env
```suggestion
			"HTTP_PROXY=" + os.Getenv("HTTP_PROXY"),
			"http_proxy"= + os.Getenv("http_proxy"),
```

cmd/yggd/main.go
subpop: Changing this line is also tangential to the PR in question here, so if we want to adjust the value of `PATH` that's set for workers, we should follow-up in a separate PR.

subpop (changes requested):

cmd/yggd/main.go
jirihnidek: Yeah, I agree that this is not related to this PR. I just wanted to ask, when I saw this line.

cmd/yggd/main.go
jirihnidek: And we should do the same for `https_proxy` environment variable. And probably also for `no_proxy`  too.

ahitacat: The issue is just with HTTP_PROXY that can be used in CGI environments, the article explains why the decision of curl.

ahitacat:
# Changelog

- Remove `HTTP_PROXY` as it can be use in CGI environments to execute scripts, that could cause some malicious problems.
- Added lowercase env variables options, `no_proxy` and `https_proxy`.

HACKING.md
```diff
@@ -59,6 +59,13 @@ export CONSUMER_ID=$(openssl x509 -in cert.pem -subject -nocert | cut -f3 -d" ")
 mosquitto_pub --host 127.0.0.1 --port 1883 --topic "yggdrasil/${CONSUMER_ID}/control/in" --message "{\"type\":\"command\",\"message_id\":\"$(uuidgen | tr -d '\n')\",\"version\":1,\"sent\":\"$(date --iso-8601=seconds --utc | tr -d '\n')\",\"content\":{\"command\":\"ping\"}}"
 ```
 
+Send a data message to `echo worker`
+```
+export CONSUMER_ID=$(openssl x509 -in cert.pem -subject -nocert | cut -f3 -d" ")
+echo "{\"type\":\"data\",\"message_id\":\"$(uuidgen | tr -d '\n')\", \"response_to\":\" \",\"version\":1,\"sent\":\"$(date --iso-8601=seconds --utc | tr -d '\n')\",\"directive\":\"echo\",\"metadata\":\"{}\",\"content\":{\"hello world\"}}" | pub -broker tcp://test.mosquitto.org:1883 -topic yggdrasil/${CONSUMER_ID}/data/in
```
subpop, line 65: In `yggdrasil-0.2`, there no other mention of `pub`. The example "ping" above uses `mosquitto_pub`. I think we should stick with consistency in the documentation, even though this is the HACKING/CONTRIBUTING guide.

```suggestion
mosquitto_pub --host 127.0.0.1 --port 1883 --topic yggdrasil/${CONSUMER_ID}/data/in --message "{\"type\":\"data\",\"message_id\":\"$(uuidgen | tr -d '\n')\", \"response_to\":\" \",\"version\":1,\"sent\":\"$(date --iso-8601=seconds --utc | tr -d '\n')\",\"directive\":\"echo\",\"metadata\":\"{}\",\"content\":{\"hello world\"}}"
```

subpop (changes requested):

HACKING.md
ahitacat: Oops sorry! Changing it.

subpop (approved):
LGTM

PR #188, opened 2023-11-14
fix: build RPM packages as part of our CI 

jirihnidek:
* The building of package was not trigger, when PR was opened.
  This commit fixes this issue


ahitacat:
This look right, but I have a small question. Now that we have COPR builds, why is needed to build the RPM in a Github action?

ahitacat (approved):

PR #189, opened 2023-11-22
feat: add Packit integration

ptoscano:
Add a basic configuration for building PRs using Packit, using EL 8/9 and supported Fedora's.

The actions are needed to do the "make srpm" way manually, which is what Copr does.

Also drop the GitHub Actions for building packages manually, as Packit covers it now (and even for more distros).


ahitacat (approved):
LGTM

PR #192, opened 2023-12-06
Use Token.WaitTimeout

subpop:
This is a backport of #165 to yggdrasil-0.2.

When connecting to a broker and publishing a message, use the
`Token.WaitTimeout` method instead of `Token.Wait`. `Token.Wait` waits
indefinitely, which can lead to situations when the cleint never
succeeds in connecting or publishing.

The timeout for each operation can be configured independently by
setting `mqtt-connect-timeout` and `mqtt-publish-timeout`. Both values
default to 30 seconds. The flags are hidden, as they should not commonly
be required to be changed by users.

Fixes: CCT-123

### DIFF
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -121,6 +121,18 @@ func main() {
 			Value:  0 * time.Second,
 			Hidden: true,
 		}),
+		altsrc.NewDurationFlag(&cli.DurationFlag{
+			Name:   "mqtt-connect-timeout",
+			Usage:  "Sets the time to wait before giving up to `DURATION` when connecting to an MQTT broker",
+			Value:  30 * time.Second,
+			Hidden: true,
+		}),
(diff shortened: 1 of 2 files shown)


ahitacat (commented):
Overall this looks good. Although I have one question. Why on mqtt.go not all the publish that use toke.Wait uses the timeout? (case Ping or cas reconnect)

subpop:
Ah, this was an oversight on my part. I wrote this patch using the `main` branch as a reference. The `main` branch has consolidated all calls to the MQTT library into the `transport.MQTTTransport` type, so that branch only had to change it in one place. For this branch, I needed to change it in more places (and just forgot). Good catch!

subpop:
Updated to include the missed uses of `Token.Wait().

ahitacat (approved):
LGTM

jirihnidek (approved):
LGTM

PR #205, opened 2024-02-05
Encode data message content as raw JSON

subpop:
When encoding a data message, the content should remain unmodified. When it is typed as a `[]byte`, it ends up getting encoded in base64. Typing it as `json.RawMessage` ensures the content is unmodified when marshaled into JSON.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,7 +79,7 @@ You should see the message logged to the output of `sub` in [Terminal
 Now publish a data message containing the string "hello" to the `echo` worker.
 
 ```
-echo "hello" | \
+echo '"hello"' | \
 go run ./cmd/yggctl generate data-message --directive echo - | \
 pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/data/in
 ```
@@ -253,7 +253,7 @@ Similarly, a data message can be published to a client using `yggctl generate`
 and `pub`.
 
 ```
-echo "hello" | \
+echo '"hello"' | \
 yggctl generate data-message --directive echo - | \
 pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/data/in
 ```
@@ -264,7 +264,7 @@ It is also possible to send data directly to a worker. This method does not publ
 a message to the MQTT broker. Instead, it calls a D-Bus method:
 
 ```bash
-echo -n "hello" | yggctl dispatch -w "echo" -
+echo -n '"hello"' | yggctl dispatch -w "echo" -
 ```
 
 # Code Guidelines
--- a/messages.go
+++ b/messages.go
@@ -118,7 +118,7 @@ type Data struct {
 	Sent       time.Time         `json:"sent"`
 	Directive  string            `json:"directive"`
 	Metadata   map[string]string `json:"metadata"`
-	Content    []byte            `json:"content"`
+	Content    json.RawMessage   `json:"content"`
 }
 
 // A WorkerMessage represents the structure of a journal entry in the


jirihnidek (changes requested):
I can see one more `[]byte` in `Response` struct. We have probably the same problem with this message too.

Isn't there any security risk of `func (m *RawMessage) UnmarshalJSON(data []byte) error { ... }`? I'm asking, because this function would be used for decoding received message.

jirihnidek:
This PR also breaks yggdrasil and everything.

Output of yggdrasil, when I try to send message to echo worker:
```
[yggd] 2024/02/06 15:24:50 /home/jhnidek/github/redhat-insights/yggdrasil/internal/work/dispatcher.go:116: received signal: &dbus.Signal{Sender:":1.14088", Path:"/com/redhat/Yggdrasil1/Worker1/echo", Name:"com.redhat.Yggdrasil1.Worker1.Event", Body:[]interface {}{0x5, "", "", map[string]string{}}, Sequence:0x7}
[yggd] 2024/02/06 15:24:50 /home/jhnidek/github/redhat-insights/yggdrasil/internal/work/dispatcher.go:120: cannot find sender: cannot call org.freedesktop.DBus.GetNameOwner: The name does not have an owner
[yggd] 2024/02/06 15:24:52 /home/jhnidek/github/redhat-insights/yggdrasil/internal/work/dispatcher.go:116: received signal: &dbus.Signal{Sender:":1.20666", Path:"/com/redhat/Yggdrasil1/Worker1/echo", Name:"com.redhat.Yggdrasil1.Worker1.Event", Body:[]interface {}{0x4, "", "", map[string]string{}}, Sequence:0xa}
[yggd] 2024/02/06 15:24:52 /home/jhnidek/github/redhat-insights/yggdrasil/cmd/yggd/client.go:192: emitted event: {Worker:echo Name:STARTED MessageID: ResponseTo: Data:map[]}
[yggd] 2024/02/06 15:25:05 /home/jhnidek/github/redhat-insights/yggdrasil/internal/transport/mqtt.go:72: cannot receive data message: cannot unmarshal data message: unexpected end of JSON input
```

Output from yggctl:
```
echo "hello" | go run ./cmd/yggctl generate data-message --directive echo - | pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/data/in
cannot marshal message: json: error calling MarshalJSON for type json.RawMessage: invalid character 'h' looking for beginning of value
exit status 1
```

subpop:
> I can see one more `[]byte` in `Response` struct. We have probably the same problem with this message too.
> 
> Isn't there any security risk of `func (m *RawMessage) UnmarshalJSON(data []byte) error { ... }`? I'm asking, because this function would be used for decoding received message.

This might not be true. The `Response` struct is a data structure that contains an optional response from Transport implementations. In the case of the HTTP Transport, it's the HTTP response body (which is a `[]byte`, not a `json.RawMessage`); the response does not have to be JSON.

jirihnidek:
> Ah yes, because the "content" field is now treated as raw JSON, it has to conform to JSON. The character string `hello` is not valid JSON. It needs to be enclosed in quotations in order to be a valid JSON snippet: `"hello"`. We could make some assumptions about incoming data and enclose it in quotations and treat it like a string, but that assumption will incorrectly treat the input `{` as a valid JSON string `"{"`.
> 
> Instead of echoing "hello", you have to echo '"hello"':
> 
> ```
> echo '"hello"' | go run ./cmd/yggctl generate data-message --directive echo -
> ```

Makes sense. We should update `CONTRIBUTING.md` file then.

subpop:
Good call. I've amended the commit to change the examples in `CONTRIBUTING.md`.

jirihnidek:
@subpop If we will change encoding of `Content` from base64 to raw, couldn't it cause any issue with compatibility? I mean, could we guarantee that we will not send raw data to system expecting base64 encoded data and vice versa?

subpop:
> @subpop If we will change encoding of `Content` from base64 to raw, couldn't it cause any issue with compatibility? I mean, could we guarantee that we will not send raw data to system expecting base64 encoded data and vice versa?

No. Encoding the content in base64 was a bug introduced when I refactored the IPC logic to D-Bus. The [protocol description](https://github.com/RedHatInsights/yggdrasil/wiki/Protocol#data-messages) of data messages does not specify a type for the "content" field. It should never have been base64-encoded to begin with.

jirihnidek (approved):
LGTM

PR #206, opened 2024-02-08
Simpler mapping of event to name

PaulWay:
Using a map rather than an if-then.

This is really just a "Hi" and me trying to get started reading through the yggdrasil code.

### DIFF
--- a/ipc/interfaces.go
+++ b/ipc/interfaces.go
@@ -53,9 +53,9 @@ const (
 	WorkerEventNameStopped WorkerEventName = 5
 )
 
-const WorkerEventNameMap := make(map[WorkerEventName]string) {
-	WorkerEventNameBegin: "BEGIN",
-	WorkerEventNameEnd: "END",
+var WorkerEventNameMap = map[WorkerEventName]string{
+	WorkerEventNameBegin:   "BEGIN",
+	WorkerEventNameEnd:     "END",
 	WorkerEventNameWorking: "WORKING",
 	WorkerEventNameStarted: "STARTED",
 	WorkerEventNameStopped: "STOPPED",
--- a/ipc/interfaces.go
+++ b/ipc/interfaces.go
@@ -53,20 +53,21 @@ const (
 	WorkerEventNameStopped WorkerEventName = 5
 )
 
+const WorkerEventNameMap := make(map[WorkerEventName]string) {
+	WorkerEventNameBegin: "BEGIN",
+	WorkerEventNameEnd: "END",
+	WorkerEventNameWorking: "WORKING",
+	WorkerEventNameStarted: "STARTED",
+	WorkerEventNameStopped: "STOPPED",
+}
+
 func (e WorkerEventName) String() string {
-	switch e {
-	case WorkerEventNameBegin:
-		return "BEGIN"
-	case WorkerEventNameEnd:
-		return "END"
-	case WorkerEventNameWorking:
-		return "WORKING"
-	case WorkerEventNameStarted:
-		return "STARTED"
-	case WorkerEventNameStopped:
-		return "STOPPED"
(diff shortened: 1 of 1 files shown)


ipc/interfaces.go
```diff
@@ -53,20 +53,21 @@ const (
 	WorkerEventNameStopped WorkerEventName = 5
 )
 
+var WorkerEventNameMap = map[WorkerEventName]string{
```
subpop, line 56: This variable is an implementation detail of the `WorkerEventName` type, so it doesn't need to be exported. In fact, it's only required as part of the `String` method on `WorkerEventName`, so it could be moved inside that function.

subpop (changes requested):
Thanks for the PR! I'm not sure how I feel about this approach over using a switch statement. I've seen the pattern of using switch statements to associate string values with const values in other Go projects, so it doesn't feel foreign to me. I don't know what we're gaining by moving the code from a switch statement to a map literal. Adding a new entry is one less line of code:

```go
WorkerEventNameFoo: "FOO",
```

instead of:

```go
case WorkerEventFoo:
    return "FOO"
```

So I'm not convinced there's much to gain by this change. If we were to change this to anything, I would look into using reflection to identify the name of the variable and see if is possible to extract the "value" component from the variable name.

ipc/interfaces.go
PaulWay: Hi @subpop thanks for your review!

I did think of putting WorkerEventNameMap in the function (I assume Go would just have it in a static symbol table, rather than creating the map every time to function was invoked).

But I guess the paradigm I've used in other code is to have code that wants to look up a event name just use the map directly, since that's what they're there for.  The function is just a convenience because we can't set a zero value for the map that isn't nil.

I don't mind either way - this was just my attempt to get into the code and learn how your git process and PR tests work :-)   If you would prefer to keep the function approach, feel free to reject this.

PaulWay: It'd also make sense to write this as a `switch`/`case` statement, because that way the Go code can work out the fastest way to evaluate the choices.  `if`/`then` is always linear.

PaulWay: And maybe this is just a reaction to that recent meme that went around of the function to determine whether a number is odd or even by having a hundred if/then statements... :grinning:

subpop: > The function is just a convenience because we can't set a zero value for the map that isn't nil.

This `String` function isn't just a convenience function. It implements the [`Stringer`](https://pkg.go.dev/fmt#Stringer) interface type for the `WorkerEventName` type. This allows instances of the type to be interpolated into format strings using the "%v" format verb. Callers shouldn't be using the string representation to determine equality of instances of the `WorkerEventName` type. The output of the `String` function is intended for inclusion in things like printed output or other visual output.

The implementation of this `WorkerEventName` type resembles the typical Go pattern for "enum" types. Since Go doesn't have the notion of enum types in the language explicitly, this kind of type alias plus constants pattern is a close equivalent.

```go
type WorkerEventName string

const WorkerEventNameBegin = "BEGIN"
const WorkerEventNameEnd = "END"

func (e WorkerEventName) String() string {
    return string(e)
}
```

In this case, `WorkerEventName`'s base type is an integer because it gets [serialized into a D-Bus message as an int](https://github.com/RedHatInsights/yggdrasil/blob/2bd2912bbf5c9893d2d30d4b6a9c2981d089050c/ipc/com.redhat.Yggdrasil1.Worker1.xml#L55-L93). Otherwise, it would be easier to use a string as the base type.

PaulWay:
Hi @subpop - thanks for your pointers there.  I'm new to Go idiom so I appreciate learning the reasons for doing it the way it is.  Will close this PR.

PR #211, opened 2024-03-08
feat: start workers explicitly

r0x0d:
Workers are now started and stopped based on the existence of a configuration file in SYSCONFDIR/yggdrasil/workers. It defines a path to the executable that will be started and managed by yggd. Conversely, the worker can be disabled by the removal of the configuration file.

Much of the code in exec.go was lifted into a new workers.go file. exec.go now containers lower level routines for starting and stopping processes. It has no awareness of the concept of a "worker". workers.go has higher-level functions that use the startProcess and stopProcess routines to start a worker process, create a PID file, wait for exit status, restart and backoff restart delays, etc.

The sample echo worker config file has been updated to include two new fields: protocol and env. protocol must be set to a supported RPC protocol (currently "grpc" is the only supported RPC protocol). Any strings in env must be in the form of a "VAR=VALUE" string, and are inserted into the worker process's environment before starting. It is forbidden to set PATH or any variable beginning with YGG_ in a worker's configuration file. Any variables by these names will be omitted when starting the worker process.

### DIFF
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -25,8 +25,12 @@ import (
 	"google.golang.org/grpc"
 )
 
-var ClientID string = ""
-var ExcludeWorkers map[string]bool = map[string]bool{}
+var (
+	ClientID                       = ""
+	SocketAddr                     = ""
+	UserAgent                      = yggdrasil.LongName + "/" + yggdrasil.Version
+	ExcludeWorkers map[string]bool = map[string]bool{}
+)
 
 func main() {
 	app := cli.NewApp()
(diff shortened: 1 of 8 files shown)


subpop:
This is a revival of #37, back-porting the changes to the `yggdrasil-0.2` branch.

cmd/yggd/main.go
```diff
@@ -25,8 +25,11 @@ import (
 	"google.golang.org/grpc"
 )
 
-var ClientID string = ""
-var ExcludeWorkers map[string]bool = map[string]bool{}
```
subpop, line 29: This patch will need to be updated to support the `--exclude-workers` flag. It was included as part of the original #37 PR, then later added to the yggdrasil-0.2 branch in #62.

r0x0d: Updated the code to reflect the ExcludeWorkers

subpop:
These changes actually introduce the explicit worker startup behavior that was part of #37 along with the configuration file. We don't actually want to bring the explicit startup behavior into yggdrasil-0.2, so I think it's best to close this PR in favor of #214, which preserves the automatic startup procedure already present in yggdrasil-0.2, while adding support for an optional configuration file that allows for specifying environment variables.

r0x0d:
Thanks, @subpop. 

Agreed, let's move with your PR instead. It looks much cleaner and better.

PR #214, opened 2024-03-13
Configurable worker environment variables

subpop:
This PR refactors the way processes are started, watched, killed, and restarted. The routines in `exec.go` were a mixture of higher-level worker process management (including restart delays, PID file management, etc.) and lower-level process start and kill routines. This PR moves the higher-level logic out of `exec.go` and into a file called `worker.go`. The routines in `exec.go` are now purely lower-level routines that start and stop processes; they have no awareness of worker configuration or process restart management (`startProcess`, `stopProcess`, `waitProcess`). The functions in `worker.go` operate on a `workerConfig` data structure to start, stop and manage workers.

A `workerConfig` structure is created by parsing an optional configuration file that must be defined in `/etc/yggdrasil/workers`. This configuration file, if present when a worker is started, is loaded. The 'env' field of the configuration file is parsed as a string slice. Elements of this slice must be in the form of a "VAR=VALUE" string, and are inserted into the worker process's environment before starting. It is forbidden to set PATH or any variable beginning with YGG_ in a worker's configuration file. Any variables by these names will be omitted when starting the worker process.

A worker's environment is automatically configured with a reasonable base PATH value (`/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`) and inherits values from its parent for: `http_proxy`, `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, and `NO_PROXY`.

### DIFF
--- a/cmd/yggd/exec.go
+++ b/cmd/yggd/exec.go
@@ -5,135 +5,84 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"path/filepath"
-	"strings"
-	"time"
 
 	"git.sr.ht/~spc/go-log"
-	"github.com/redhatinsights/yggdrasil"
 )
 
-func startProcess(file string, env []string, delay time.Duration, died chan int) {
+type processStartedFunc func(pid int, stdout, stderr io.ReadCloser)
+type processStoppedFunc func(pid int, state *os.ProcessState)
+
+// startProcess executes file, setting up the environment using the provided env
+// values. If the function parameter started is not nil, it is invoked on a
+// goroutine after the process has been started.
+func startProcess(
+	file string,
+	args []string,
+	env []string,
+	started processStartedFunc,
+) error {
 	if _, err := os.Stat(file); os.IsNotExist(err) {
-		log.Warnf("cannot start worker: %v", err)
-		return
+		return fmt.Errorf("cannot find file: %v", err)
 	}
 
-	cmd := exec.Command(file)
+	cmd := exec.Command(file, args...)
 	cmd.Env = env
-
-	if delay < 0 {
-		log.Errorf("failed to start worker '%v' too many times", file)
-		return
-	}
-
-	if delay > 0 {
-		log.Tracef("delaying worker start for %v...", delay)
-		time.Sleep(delay)
-	}
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
-		log.Errorf("cannot connect to stdout: %v", err)
-		return
+		return fmt.Errorf("cannot connect to stdout: %v", err)
 	}
 
 	stderr, err := cmd.StderrPipe()
 	if err != nil {
-		log.Errorf("cannot connect to stderr: %v", err)
-		return
+		return fmt.Errorf("cannot connect to stderr: %v", err)
 	}
 
 	if err := cmd.Start(); err != nil {
-		log.Errorf("cannot start worker: %v: %v", file, err)
-		return
+		return fmt.Errorf("cannot start worker: %v: %v", file, err)
+
 	}
 	log.Debugf("started process: %v", cmd.Process.Pid)
 
-	go func() {
-		for {
-			buf := make([]byte, 4096)
-			n, err := stdout.Read(buf)
-			if n > 0 {
-				log.Tracef("[%v] %v", file, strings.TrimRight(string(buf), "\n\x00"))
-			}
-			if err != nil {
-				switch err {
-				case io.EOF:
-					log.Debugf("%v stdout reached EOF: %v", file, err)
-					return
-				default:
-					log.Errorf("cannot read from stdout: %v", err)
-					continue
-				}
-			}
-		}
-	}()
-
-	go func() {
-		for {
-			buf := make([]byte, 4096)
-			n, err := stderr.Read(buf)
-			if n > 0 {
-				log.Errorf("[%v] %v", file, strings.TrimRight(string(buf), "\n\x00"))
-			}
-			if err != nil {
-				switch err {
-				case io.EOF:
-					log.Debugf("%v stderr reached EOF: %v", file, err)
-					return
-				default:
-					log.Errorf("cannot read from stderr: %v", err)
-					continue
-				}
-			}
-		}
-	}()
-
-	pidDirPath := filepath.Join(yggdrasil.LocalstateDir, "run", yggdrasil.LongName, "workers")
-
-	if err := os.MkdirAll(pidDirPath, 0755); err != nil {
-		log.Errorf("cannot create directory: %v", err)
-		return
+	if started != nil {
+		go started(cmd.Process.Pid, stdout, stderr)
 	}
 
-	if err := os.WriteFile(filepath.Join(pidDirPath, filepath.Base(file)+".pid"), []byte(fmt.Sprintf("%v", cmd.Process.Pid)), 0644); err != nil {
-		log.Errorf("cannot write to file: %v", err)
-		return
-	}
-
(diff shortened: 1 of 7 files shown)


cmd/yggd/worker.go
```diff
@@ -2,60 +2,273 @@ package main
...
+				return fmt.Errorf("cannot stop worker: %w", err)
+			}
 		}
 	}
 
 	return nil
 }
 
-func watchWorkerDir(dir string, env []string, died chan int) {
+func watchWorkerDir(died chan int) {
```
jirihnidek, line 266: I would add some doc string here.

cmd/yggd/worker.go
```diff
@@ -65,27 +278,52 @@ func watchWorkerDir(dir string, env []string, died chan int) {
...
+			if err := stopWorker(workerName); err != nil {
+				log.Errorf("cannot stop worker: %v", err)
 				continue
 			}
 		}
 	}
 }
+
+func validEnvVar(val string) bool {
+	for _, pattern := range []string{"PATH=.*", "YGG_.*=.*"} {
```
jirihnidek, line 317: What about `LD_PRELOAD` or `PYTHONPATH`?

cmd/yggd/worker.go
```diff
@@ -65,27 +278,52 @@ func watchWorkerDir(dir string, env []string, died chan int) {
...
+		r := regexp.MustCompile(pattern)
+		if r.Match([]byte(val)) {
```
jirihnidek, line 319: I would print some warning log that given environment variable was filtered out.

cmd/yggd/exec.go
```diff
@@ -5,135 +5,80 @@ import (
...
 	}
 
-	go startProcess(cmd.Path, cmd.Env, delay, died)
+	return nil
 }
 
-func killProcess(pid int) error {
-	process, err := os.FindProcess(int(pid))
+// stopProcess finds a process with the given pid and kills it.
+func stopProcess(pid int) error {
```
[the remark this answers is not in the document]
subpop: I felt like "start/kill" doesn't follow a naming convention, so I chose to rename `killProcess` to `stopProcess`. Now it follows the "start/stop" convention.

cmd/yggd/exec.go
```diff
@@ -5,135 +5,80 @@ import (
...
 
-func startProcess(file string, env []string, delay time.Duration, died chan int) {
+// startProcess executes file, setting up the environment using the provided env
+// values. If started is not nil, it is invoked on a goroutine after the process
+// has been started.
+func startProcess(
+	file string,
+	args []string,
+	env []string,
+	started func(pid int, stdout io.ReadCloser, stderr io.ReadCloser),
```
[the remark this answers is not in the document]
subpop: Good point. I spent some time researching conventions around naming function parameters and couldn't find any definitive sources. There's lots of opinion on the Internet (no surprise). I try to follow the Go standards as much as possible, and in the absence of documentation by the Go project (as is the case here), I look at the Go standard library for inspiration.

* The [net/http](https://pkg.go.dev/net/http) package declares a function called [HandleFunc](https://pkg.go.dev/net/http#HandleFunc). This function takes a function as a parameter. The parameter is named "handler".
* The [sort](https://pkg.go.dev/sort) package includes a number of examples of functions that take functions as parameters too. They also use short, evocative names like "less" and "cmp".
* The [strings.Map](https://pkg.go.dev/strings#Map) function names a parameter function "mapping".
* The [Effective Go section on names](https://go.dev/doc/effective_go#names) puts emphasis on keeping names short and concise.

I interpret this to mean that there is no Go convention around using a parameter name to hint at its type. We certainly don't do this for other parameter types; a function definition of `func validate(nameString string, uppercaseBool bool)` would not be following convention. To me, a function that's also a function parameter should have a name that implies action. Hence "started" feels like a name that suggests action was taken, since it uses the past conjugation of the verb "to start".

I can rewrite the comment documentation to more clearly use the parameter name as a value and not as a verb, but I'm pretty sure that `started` is following about as close of a convention as I can find.

cmd/yggd/exec.go
```diff
@@ -5,135 +5,80 @@ import (
...
-		delay += 5 * time.Second
+	log.Debugf("waiting for process: %v", process.Pid)
+	state, err := process.Wait()
+	if err != nil {
+		return fmt.Errorf("process %v exited with error: %v", process.Pid, err)
 	}
-	if delay >= time.Duration(30*time.Second) {
-		delay = -1
+
+	if died != nil {
```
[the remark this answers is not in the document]
subpop: Leading from my findings in the thread about `started`, I think I'd prefer to avoid the `Callback` suffix, but I agree that "died" is not the right name; "exited" or "stopped" is probably a better name.

cmd/yggd/worker.go
subpop: Good idea.

cmd/yggd/worker.go
subpop: Ah, yea, those are probably worth keeping out, but if we start to exclude every language's library search path, where will it end? Maybe it's better to prefix each inserted variable with the name of the worker, thereby permitting all variables, but with running the risk of clobbering conventional ones like `LD_PRELOAD`?

cmd/yggd/util.go
```diff
@@ -53,3 +53,8 @@ func parseCertCN(filename string) (string, error) {
 	}
 	return cert.Subject.CommonName, nil
 }
+
+func fileExists(file string) bool {
```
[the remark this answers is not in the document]
subpop: I disagree. One of the tennants of Go is concise naming patterns. In the case of
this function, it is actually small enough that `file` could be renamed to `f`
and the meaning is still derivable. See [Local
Variable](https://go.dev/talks/2014/names.slide#6) in the "What's in a name?"
presentation.

cmd/yggd/worker.go
```diff
@@ -2,60 +2,273 @@ package main
...
+					case io.EOF:
+						log.Debugf("%v stderr reached EOF: %v", config.program, err)
+						return
+					default:
+						log.Errorf("cannot read from stderr: %v", err)
+						continue
+					}
+				}
+			}
+		}()
```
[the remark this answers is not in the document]
subpop: Good point!

cmd/yggd/worker.go
subpop: Oh. That's an idea. Are you suggesting a per-worker list of forbidden environment variables? Or a single, global list of forbidden environment variables that's kept in `/etc/rhc/config.toml`?

jirihnidek: I think that It would make sense to have both of them. One global list containing e.g. `PATH`, `LD_PRELOAD`, `YGG_.*` environment variables, and then per-worker list. When worker is written in Python, then it makes sense to extend global list with `PYTHONPATH` from per-worker list for given worker. It would have less sense to have such variable for worker written in Go or C.

On the other side, I'm not sure if it is worth adding such new complexity to this branch?

subpop: Yea. Since this isn't a configuration that's going to be carried forward into yggdrasil 0.3+, I'd rather not add a configuration value that we aren't supporting in the future. Can we leave it as a hard-coded list of `PATH` and `YGG_*` and call that good enough for now?

jirihnidek: Yeah, I think there is small risk from environment variables, because yggdrasil is typically started using systemd service. Only root can modify environment variables in this service.

jirihnidek (approved):
LGTM

Issue #210, opened 2024-02-29
Support hyphenated directive names with a warning


subpop:
The D-Bus object name specification does not permit hyphens (dashes) in the name. For example, `com.redhat.Yggdrasil1.Worker1.rhc-worker-playbook` is not valid. Underscores are valid. Workers using yggdrasil >= 0.3 that include hyphens in their directive string will need to update to using underscores. To support existing directives, yggdrasil should permit hyphens, but should log a warning.

~In the `worker` package, [this error](https://github.com/RedHatInsights/yggdrasil/blob/c3d9e18c2755a33c1936d9717996f95d3475a840/worker/worker.go#L49-L52) should be replaced with a warning, and the hyphens should be replaced with underscores.~

In the `internal/work` package, before [looking for a worker](https://github.com/RedHatInsights/yggdrasil/blob/c3d9e18c2755a33c1936d9717996f95d3475a840/internal/work/dispatcher.go#L196), the hyphens should be replaced with underscores.

~In `libygg`, the [`ygg_worker_connect` method](https://github.com/RedHatInsights/libygg/blob/fe5b1135dd8891516598c0125f23dd54b7e86ad7/src/ygg-worker.c#L511-L519) should log a warning and replace hyphens with underscores.~

PR #216, opened 2024-03-21
Allow hyphenated names, with a warning

subpop:
The D-Bus name specification does not allow hyphens (-) in names or paths. Workers should really switch to using compliant names. However, in order to maintain compatibility with existing services, when a message is received for a worker, the name is passed through a scrubbing function that cleans up the name. As of this PR, this scrub function replaces hyphens with underscores and returns an error. This error is then logged. These log messages should hopefully encourage service authors to switch to using compatible names.

Closes: #210

### DIFF
--- a/internal/work/dbus.go
+++ b/internal/work/dbus.go
@@ -1,6 +1,11 @@
 package work
 
-import "github.com/godbus/dbus/v5"
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/godbus/dbus/v5"
+)
 
 func NewDBusError(name string, body ...string) *dbus.Error {
 	e := dbus.Error{}
@@ -9,4 +14,17 @@ func NewDBusError(name string, body ...string) *dbus.Error {
 		e.Body = append(e.Body, v)
 	}
 	return &e
+}
+
+// ScrubName cleans up invalid bus names to ensure D-Bus name specification
+// conformance. An error is returned along with the scrubbed value if the name
+// contained invalid characters.
+func ScrubName(name string) (string, error) {
+	r := regexp.MustCompile("-")
+	if r.Match([]byte(name)) {
+		return string(
+			r.ReplaceAll([]byte(name), []byte("_")),
+		), fmt.Errorf("invalid worker name: %v", name)
+	}
+	return name, nil
 }
--- a/internal/work/dbus_test.go
+++ b/internal/work/dbus_test.go
@@ -1,0 +1,61 @@
+package work
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+)
+
+func TestScrubName(t *testing.T) {
+	tests := []struct {
+		description string
+		input       string
+		want        string
+		wantError   error
+	}{
+		{
+			description: "no hyphens",
+			input:       "alpha",
+			want:        "alpha",
+		},
+		{
+			description: "single hyphen",
+			input:       "alpha-bravo",
+			want:        "alpha_bravo",
+			wantError:   cmpopts.AnyError,
+		},
+		{
+			description: "multiple hyphens",
+			input:       "alpha-bravo-charlie",
+			want:        "alpha_bravo_charlie",
+			wantError:   cmpopts.AnyError,
+		},
+		{
+			description: "consecutive hyphens",
+			input:       "alpha--bravo",
+			want:        "alpha__bravo",
+			wantError:   cmpopts.AnyError,
+		},
+		{
+			description: "mixed",
+			input:       "alpha_bravo-charlie",
+			want:        "alpha_bravo_charlie",
+			wantError:   cmpopts.AnyError,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			got, err := ScrubName(test.input)
+
+			if !cmp.Equal(got, test.want) {
+				t.Errorf("%v", cmp.Diff(got, test.want))
+			}
+
+			if !cmp.Equal(err, test.wantError, cmpopts.EquateErrors()) {
+				t.Errorf("%#v != %#v", err, test.wantError)
+			}
+		})
+	}
+}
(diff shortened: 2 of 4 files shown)


internal/work/errors.go
```diff
@@ -33,3 +33,15 @@ func newStringMapTypeConversionError(parameter interface{}) typeConversionError
 		t: reflect.TypeOf(map[string]string{}),
 	}
 }
+
+var ErrInvalidWorkerName WorkerError = "invalid worker name"
+
+type WorkerError string
```
jirihnidek, line 39: The name of error looks to generic, but it is related only to invalid name directive (not name of worker). The name of worker could be correct, but this error should be still raised.

I believe that code could be simplified. I propose something like this:

```go
type InvalidDirectiveNameError struct {
	InvalidName string
}

func (e InvalidDirectiveNameError) Error() string {
	return fmt.Sprintf("invalid directive name: %s", e.InvalidName)
}
```

internal/work/dbus_test.go
```diff
@@ -0,0 +1,61 @@
...
+		{
+			description: "no hyphens",
+			input:       "alpha",
+			want:        "alpha",
+		},
+		{
+			description: "single hyphen",
+			input:       "alpha-bravo",
+			want:        "alpha_bravo",
+			wantError:   ErrInvalidWorkerName,
```
jirihnidek, line 26: This should be changed to something like `InvalidDirectiveNameError{"alpha-bravo"}` and other cases too.

internal/work/dbus.go
```diff
@@ -10,3 +15,16 @@ func NewDBusError(name string, body ...string) *dbus.Error {
...
+
+// ScrubName cleans up invalid bus names to ensure D-Bus name specification
+// conformance. An error is returned along with the scrubbed value if the name
+// contained invalid characters.
+func ScrubName(name string) (string, error) {
+	r := regexp.MustCompile("-")
+	if r.Match([]byte(name)) {
+		return string(
+			r.ReplaceAll([]byte(name), []byte("_")),
+		), fmt.Errorf("%w: %v", ErrInvalidWorkerName, name)
```
jirihnidek, line 27: This could be simplified to something like:

```go
	        return string(
			r.ReplaceAll([]byte(name), []byte("_")),
		), InvalidDirectiveNameError{name}
```

jirihnidek (changes requested):
I have several comments, suggestions and requests. More details in comments.

worker/worker.go
```diff
@@ -48,7 +48,8 @@ func NewWorker(
 ) (*Worker, error) {
 	r := regexp.MustCompile("-")
 	if r.Match([]byte(directive)) {
-		return nil, fmt.Errorf("invalid directive '%v'", directive)
+		log.Warnf("invalid directive '%v': replacing hyphens with underscores", directive)
+		directive = string(r.ReplaceAll([]byte(directive), []byte("_")))
```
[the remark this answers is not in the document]
subpop: Ahh good point. We should only maintain backwards-compatibility in the inbound-from-service direction. We should force worker authors to conform to the new pattern.

internal/work/dispatcher.go
```diff
@@ -193,6 +193,12 @@ func (d *Dispatcher) Connect() error {
 }
 
 func (d *Dispatcher) Dispatch(data yggdrasil.Data) error {
+	var err error
+	data.Directive, err = ScrubName(data.Directive)
+	if err != nil {
+		log.Warn(err)
```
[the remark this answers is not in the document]
subpop: That's true. I set it at a higher level in the hopes that a user would see it and it would encourage service developers to fix their directive names. That's probably wishful thinking though.

internal/work/dbus_test.go
subpop: This breaks the error test though, because the two errors are not the same. I had an implementation similar to what you're suggesting and the error tests didn't pass. I'm not *thrilled* with this solution either though. I did this only to allow for testing of the error cases. Maybe it's not worth defining a separate error type.

internal/work/dispatcher.go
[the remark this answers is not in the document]
jirihnidek: Will be service developers able to change directive name? They will have to support old and new workers, right. Will they be able to distinguish, what yggdrasil/rhcd version is on the other side? The yggdrasil.service still send following message during connecting:

```json
{
    "type": "connection-status",
    "message_id": "41c09a98-cbe5-4b79-82a0-63493a617bfc",
    "response_to": "",
    "version": 1,
    "sent": "2024-03-22T09:24:33.864006624+01:00",
    "content": {
        "canonical_facts": null,
        "dispatchers": {},
        "state": "online"
    }
}
```

The version is still the same (1). Should not we bump up this version?

internal/work/dbus_test.go
jirihnidek: Was `Is(target error)` function implemented correctly? When I change it to something like this:

```go
func (e WorkerError) Is(target error) bool {
	return e.Error() == target.Error()
}
```

then all tests passes with my suggested changes.

subpop: Comparing the values of the `Error` function is considered bad practice. This is discussed in detail in [an issue](https://github.com/google/go-cmp/issues/289) on the go-cmp project.

internal/work/dispatcher.go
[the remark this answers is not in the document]
subpop: Yes, service developers can change the directive name, but they need to be aware of old clients running the old directive name. They can distinguish between client versions because new clients (yggdrasil 0.3 and newer) [include the client version in the connection-status message](https://github.com/RedHatInsights/yggdrasil/commit/512591a68671ca957f85ed68b773efbfdaeecc45). The version field is currently unused, but was added for situations like this where we need to break protocol conformance. I don't think that this is the case here though; examining the [`content.client_version` field](https://github.com/RedHatInsights/yggdrasil/wiki/Protocol#connection-status) should be sufficient.

jirihnidek (approved):
LGTM

PR #223, opened 2024-04-01
Explict HTTP request method "Request"

subpop:
Add a new "com.redhat.Yggdrasil1.Dispatcher1.Request" D-Bus method. This
method transmits data using an HTTP request. The existing "Transmit"
method now exclusively transmits data over the configured transport
(typically MQTT). Workers no longer need to set "RemoteContent" to true
or false. A worker that previously relied on the "RemoteContent" feature
will now receive the URL directly as the message body. It is the
worker's responsibility to unpack the URL and get the URL body. The
"Request" method is available to provide an authenticated HTTP client to
make HTTP requests easier. Workers are free to choose either of the
"Transmit" or the "Request" method to return data; they are no longer
forced to use one method or the other.


README.md
```diff
@@ -126,11 +126,17 @@ the bus:
...
+is left to worker discretion, though it is recommended that messages larger than
+a few megabytes be sent using HTTP.
 
 Package `worker` implements the above requirements implicitly, enabling workers
 to be written without needing to worry about much of the D-Bus requirements
 outlined above.
 
-See `worker/echo` for a reference implementation of a worker program.
+See `worker/echo` for a reference implementation of a worker program that uses
+"Transmit". See `worker/http` for a demonstrative worker that uses "Request".
```
jirihnidek, line 142: I would add somewhere example how to send MQTT message to new http worker. Example would be:

```
echo '{"method": "GET", "url": "http://httpbin.org/get", "headers": {"User-Agent": "yggdrasil-http-worker"}}' | \ 
    go run ./cmd/yggctl generate data-message --directive http - | \
    pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/data/in
```

internal/work/dispatcher.go
```diff
@@ -297,99 +259,96 @@ func (d *Dispatcher) Transmit(
...
+		},
+		Resp: ch,
+	}
+
+	select {
+	case resp := <-ch:
+		responseCode = resp.Code
+		responseMetadata = resp.Metadata
+		responseData = resp.Data
+	case <-time.After(1 * time.Second):
```
jirihnidek, line 285: Why is timeout only one second? I think that this is too low value. Could it be configured somehow? Could it be e.g. part of MQTT message sent to worker? If the `timeout` would not be part of MQTT message, then some default value would be used.

worker/http/main.go
```diff
@@ -0,0 +1,101 @@
...
+
+	for i := 0; i < loops; i++ {
+		responseCode, responseHeaders, responseBody, err := w.Request(
+			message.Method,
+			message.URL,
+			message.Headers,
+			message.Body,
+		)
+		if err != nil {
+			return fmt.Errorf("cannot call Request: %v", err)
```
jirihnidek, line 62: The message is not accurate it was possible to call `Request`, but it failed.

worker/http/README.md
```diff
@@ -0,0 +1,51 @@
...
+```json
+{
+    "type": "object",
+    "properties": {
+        "method": { "type": "string" },
+        "url": { "type": "string" },
+        "headers": {
+            "type": "object",
+            "patternProperties": {
+                "^.*$": { "type": "string" },
```
jirihnidek, line 28: Extra comma.

worker/http/README.md
```diff
@@ -0,0 +1,51 @@
...
+            },
+        },
+        "body": { "type": "string" },
```
jirihnidek, line 31: Extra comma.

worker/http/README.md
```diff
@@ -0,0 +1,51 @@
...
+{
+    "type": "object",
+    "properties": {
+        "method": { "type": "string" },
+        "url": { "type": "string" },
+        "headers": {
+            "type": "object",
+            "patternProperties": {
+                "^.*$": { "type": "string" },
+            },
```
jirihnidek, line 29: Extra comma.

worker/http/README.md
```diff
@@ -0,0 +1,51 @@
...
+default, the worker will attempt to connect to an appropriate D-Bus (session or
+system), depending on whether `DBUS_SESSION_BUS_ADDRESS` is defined.
+
+## Usage
+
+The worker expects to receive messages in JSON conforming to the following JSON
+schema definition.
+
+### Schema
+```json
```
jirihnidek, line 19: This is not valid JSON document...

jirihnidek (commented):
I know that this is still draft, but I have few comments.

jirihnidek:
I also tried to test it with https server and it did not work. Let's try this:

```
echo '{"method": "GET", "url": "https://httpbin.org/get", "headers": {"User-Agent": "yggdrasil-http-worker"}}' | \
    go run ./cmd/yggctl generate data-message --directive http - | \
    pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/data/in
```

README.md
subpop: I feel like this kind of detail belongs more in the CONTRIBUTING document,
rather than the README. The README is intended for end users more than
contributors. CONTRIBUTING contains existing instructions on how to interact
with workers; maybe I can add a section in that document on how to interact with
the `http` worker.

internal/work/dispatcher.go
subpop: With the separation of `Transmit` and `Request`, this timeout now applies only
to the configured data transport's implementation of `Tx`. This timeout is
waiting for a response to the underlying transport `Tx` call. In practice, this
happens instantaneously with the MQTT transport because the MQTT transport does
not support responses. The HTTP transport does respond to `Transmit` requests
with the response of the underlying HTTP request. This is an unfortunate side
effect of supporting both MQTT and HTTP as data transports.

I am not sure whether it is worth exposing this as a configurable value. Maybe
we could re-use the existing `config.HTTPTimeout` configuration value?

README.md
subpop: I added a section to the `http` worker README.

PR #225, opened 2024-04-03
feat(echo): set log level according to env

subpop:
Set the log level according to the value of YGG_LOG_LEVEL.
Print the worker's environment when run at level Debug or higher.

### DIFF
--- a/worker/echo/main.go
+++ b/worker/echo/main.go
@@ -16,6 +16,16 @@ import (
 var yggdDispatchSocketAddr string
 
 func main() {
+	// Set log level inherited from the environment
+	if logLevel, ok := os.LookupEnv("YGG_LOG_LEVEL"); ok {
+		level, err := log.ParseLevel(logLevel)
+		if err != nil {
+			log.Fatalf("cannot parse log level: %v", err)
+		}
+		log.SetLevel(level)
+	}
+	log.Debugf("current environment: %v", os.Environ())
+
 	// Get initialization values from the environment.
 	var ok bool
 	yggdDispatchSocketAddr, ok = os.LookupEnv("YGG_SOCKET_ADDR")
--- a/cmd/yggd/worker.go
+++ b/cmd/yggd/worker.go
@@ -47,6 +47,7 @@ func loadWorkerConfig(file string) (*workerConfig, error) {
 	if err := toml.Unmarshal(data, &config); err != nil {
 		return nil, fmt.Errorf("cannot load config: %w", err)
 	}
+	log.Debugf("loaded worker config: %v", config)
 
 	return &config, nil
 }
@@ -98,6 +99,7 @@ func startWorker(config workerConfig, started workerStartedFunc, stopped workerS
 		time.Sleep(config.delay)
 	}
 
+	log.Debugf("starting worker %v with environment: %v", config.program, env)
 	err := startProcess(config.program, nil, env, func(pid int, stdout, stderr io.ReadCloser) {
 		log.Infof("started worker: %v", config.program)
 


jirihnidek:
I had some issues with testing `echo` worker for any branch forked from `yggdrasil-0.2`, because instructions in `HACKING.md` are not correct.

1) Command for getting current consumer UUID is not correct:

```
export CONSUMER_ID=$(openssl x509 -in cert.pem -subject -nocert | cut -f3 -d" ")
```

because it the path to `cert.pem` is relative path. It should be absolute path (`/etc/pki/consumer/cert.pem`). If this is fixed, then environment variable contains `,` at the end. The command has to be run using `sudo`.

When fixed, then it should be something like this:

```
export CONSUMER_ID=$(sudo openssl x509 -in /etc/pki/consumer/cert.pem -subject -nocert | cut -f3 -d" " | cut -c-36)
```

2) Command for sending message is wrong. It should be probably something like this:

```
mosquitto_pub --host 127.0.0.1 --port 1883 --topic yggdrasil/${CONSUMER_ID}/data/in --message "{\"type\":\"data\",\"message_id\":\"$(uuidgen | tr -d '\n')\", \"response_to\":\" \",\"version\":1,\"sent\":\"$(date --iso-8601=seconds --utc | tr -d '\n')\",\"directive\":\"echo\",\"content\":\"hello world\"}"
```

3) Installing `echo` worker using `sudo install -D -m 755 echo /usr/libexec/yggdrasil/echo-worker` is not what you want, because when yggd tries to dispatch message, then it says that there is no echo worker...

```
[yggd] 2024/04/04 17:24:54 /home/jhnidek/github/redhat-insights/yggdrasil-0.2/cmd/yggd/mqtt.go:18: received a message on topic yggdrasil/ca38df52-c56a-47af-96dc-2416b819fba4/data/in
[yggd] 2024/04/04 17:24:54 /home/jhnidek/github/redhat-insights/yggdrasil-0.2/cmd/yggd/mqtt.go:25: message: 363a07d1-eedc-4f3f-8bf6-94223b61005f
[yggd] 2024/04/04 17:24:54 /home/jhnidek/github/redhat-insights/yggdrasil-0.2/cmd/yggd/grpc.go:125: cannot route message to directive: echo
```

The echo worker should be installed using:

```
sudo install -D -m 755 echo /usr/local/libexec/yggdrasil/echo-worker
```

When echo worker is installed to `/usr/local/libexec/yggdrasil/echo-worker`, then it seems that it works as expected.

subpop:
Are these changes introduced by this PR? If not, can you submit a PR to update HACKING?

jirihnidek:
> Are these changes introduced by this PR? If not, can you submit a PR to update HACKING?

Sure I will create PR... I just wanted to add some notes about it somewhere. And the comment was the right place for the case somebody else want to test it.

jirihnidek:
The PR with updated instructions: https://github.com/RedHatInsights/yggdrasil/pull/227

worker/echo/main.go
```diff
@@ -16,6 +16,16 @@ import (
 var yggdDispatchSocketAddr string
 
 func main() {
+	// Set log level inherited from the environment
+	if logLevel, ok := os.LookupEnv("YGG_LOG_LEVEL"); ok {
+		level, err := log.ParseLevel(logLevel)
+		if err != nil {
+			log.Fatalf("cannot parse log level: %v", err)
```
jirihnidek, line 23: When log level could not be parsed, is it good idea to set this wrong log level. Shouldn't we skip skip setting log level in this case?

worker/echo/main.go
```diff
@@ -16,6 +16,16 @@ import (
...
+		}
+		log.SetLevel(level)
+	}
+	log.Debug(os.Environ())
```
jirihnidek, line 27: Shouldn't be this debug print introduced with some log message? Something like: "current environment variables"?

jirihnidek (changes requested):
Overall good. I have only two questions/suggestions.

worker/echo/main.go
subpop: `log.Fatal` will print the message string and then exit the program, so here we are not setting an invalid level. The program is exiting.

worker/echo/main.go
subpop: Yes, this is a good approach.

worker/echo/main.go
jirihnidek: Oh, I didn't know that. It is then OK.

jirihnidek (approved):
LGTM

PR #236, opened 2024-05-13
Add initial support for tmt tests

subpop:
Add a basic tmt plan and a single smoke test. This test runs yggd directly out of the source tree. It does not compile an RPM and test it.


jirihnidek (commented):
I think that it is good start, but your PR did not trigger any test on testing farm. Why didn't you decide to use Packit integration as `rhc` uses: https://github.com/RedHatInsights/rhc/pull/107 ?

subpop:
> Why didn't you decide to use Packit integration as `rhc` uses: [RedHatInsights/rhc#107](https://github.com/RedHatInsights/rhc/pull/107) ?

Because I forgot. I'll add those packit jobs to this PR too.

subpop:
CentOS Stream 9 is failing because the spec file checks for the presence of the go-srpm-macros. On CentOS Stream 8, that check fails, and the spec file compiles things manually. On CentOS Stream 9, the macro exists. This means the spec file ends up creating an additional source package `golang-github-redhatinsights-yggdrasil-devel`. This package exists to support the Fedora-style Go source package structure. `tmt` installs all artifacts from the COPR build by default, including this source package, but it fails. This source package depends on other golang packages, none of which exist in CentOS Stream 9, so the package installation fails.

We can solve this in two ways:

1) Tell `tmt` to not install the `golang-github-redhatinsights-yggdrasil-devel` package. This is okay to do; that RPM is not required to test the functionality of the software. It's just a by-product of the go-srpm-macros. I don't know if this is even possible, or how to do it.
2) Update the spec file to only use go-srpm-macros on Fedora. I'm inclined to take this approach. I know that CentOS is not adopting Fedora's golang packaging model, so we are expected to use vendored Go packages.

jirihnidek:
>     2. Update the spec file to only use go-srpm-macros on Fedora. I'm inclined to take this approach. I know that CentOS is not adopting Fedora's golang packaging model, so we are expected to use vendored Go packages.

I agree.

jirihnidek (approved):
LGTM

PR #237, opened 2024-05-22
Add dist workflow

subpop:
With this workflow, when a new git tag is created using the semantic versioning syntax (vX.Y.Z), this workflow checkout the tag from git, generates a `yggdrasil-X.Y.Z.tar.xz` and attaches it as an asset to the release.

[Here's an example](https://github.com/subpop/yggdrasil/actions/runs/9197368705/job/25297639408) of the workflow running on my fork.

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -1,0 +1,20 @@
+name: dist
+
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


jirihnidek:
It looks good, but given link to your fork reference following link: https://github.com/subpop/yggdrasil/releases/tag/v0.4.2, but it is not valid link. It seems that reason is that such tag does not exists ATM.

subpop:
Oh, yea, I've been creating and deleting that tag to test the trigger conditions of the action. I didn't see a point to keeping around a release tag in my fork that isn't a real release, so I deleted it after testing.

jirihnidek (approved):
LGTM

PR #239, opened 2024-07-12
Add distributed SELinux policy

subpop:
Add a build option to enable building and installing an SELinux policy.


subpop:
I added `files_watch_etc_files(yggd_t)` to allow watching files in /etc.

subpop:
I haven't considered how far back in compatibility we want to go yet, but not supporting EL8 is very likely.

jirihnidek:
@subpop Is it still draft? Or do you need final review and review from selinux team?

selinux/yggdrasil.te
```diff
@@ -0,0 +1,41 @@
...
+allow yggd_t self:capability { setgid setuid };
+allow yggd_t self:fifo_file rw_fifo_file_perms;
+allow yggd_t self:unix_stream_socket create_stream_socket_perms;
+allow yggd_t self:unix_dgram_socket { connect create getattr };
+
+domain_use_interactive_fds(yggd_t)
+
+files_read_etc_files(yggd_t)
+
+files_watch_etc_files(yggd_t)
```
jirihnidek, line 27: I have just note that is not related to this case. We had to consider use case for rhsm.service, when some customers use NFS for sharing `/etc` directory and it wasn't possible to use i-notify. We had to implement regular polling as fallback solution. I hope such request will not pop up for any client tool anymore.

subpop: That's alarming. Machine-specific information goes into `/etc`. Sharing it over NFS seems unsafe. I feel like that's a condition that we should not support unless we're explicitly required to. That introduces a lot of complexity unnecessarily.

PR #248, opened 2024-08-05
feat(yggctl): add generate worker-data command

subpop:
The generate worker-data command creates files relevant to D-Bus and systemd in order for a worker to communicate with yggd. This is intended to make worker development easier; if a worker has a relatively standard startup and policy configuration, it can be configured to invoke this subcommand instead of keeping the files in its own source repository.

The files can either be installed directly by specifying the --install flag, or they can be written to a given output directory.

The --user, --program and --name flags must be included in the invocation. For example:

```
yggctl generate worker-data \
  --name echo \
  --program /usr/libexec/yggdrasil/echo \
  --user $(pkgconf yggdrasil --variable worker_user \
  --install
```

Card ID: CCT-680

### DIFF
--- a/cmd/yggctl/actions.go
+++ b/cmd/yggctl/actions.go
@@ -7,12 +7,14 @@ import (
 	"io"
 	"log"
 	"os"
+	"path/filepath"
 	"text/tabwriter"
 	"text/template"
 
 	"github.com/godbus/dbus/v5"
 	"github.com/google/uuid"
 	"github.com/redhatinsights/yggdrasil"
+	"github.com/redhatinsights/yggdrasil/internal/constants"
 	"github.com/redhatinsights/yggdrasil/ipc"
 	"github.com/urfave/cli/v2"
 )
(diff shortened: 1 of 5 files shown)


cmd/yggctl/main.go
```diff
@@ -97,6 +98,58 @@ from stdin.`,
...
+							Aliases:  []string{"p"},
+							Usage:    "set the worker program to `PATH`",
+							Required: true,
+						},
+						&cli.StringFlag{
+							Name:     "user",
+							Aliases:  []string{"u"},
+							Usage:    "set the worker user to `USER`",
+							Required: true,
+						},
```
jirihnidek, line 135: What about adding arbitrary `StringFlag` "group"? because user could be different in some case from the group.

cmd/yggctl/main.go
```diff
@@ -97,6 +98,58 @@ from stdin.`,
...
+						if ctx.String("output") == "" && !ctx.Bool("install") {
+							return cli.Exit(
+								"error: you must specify either --install or --output",
+								1,
+							)
+						}
+
+						if strings.Contains(ctx.String("name"), " -") {
+							return cli.Exit("'name' cannot contain spaces or dashes", 1)
+						}
```
jirihnidek, line 147: We could also check if given user is valid user here. E.g. the username cannot contain spaces. Not sure if we should check if the given user is valid user.

cmd/yggctl/actions.go
```diff
@@ -306,6 +308,89 @@ func listenAction(ctx *cli.Context) error {
 	return nil
 }
 
+func generateWorkerDataAction(ctx *cli.Context) error {
```
jirihnidek, line 311: I would add doc string to this new function.

jirihnidek (changes requested):
I like the idea. I have only few comments and suggestions.

cmd/yggctl/main.go
subpop: Ah yea, good idea.

cmd/yggctl/main.go
subpop: Yes. I didn't want to check if the user was an active user on the running system, since this could be used during build time to generate files that are destined to be installed on another system. However, we can validate the inputs better. 👍

jirihnidek (approved):
LGTM

Issue #135, opened 2023-06-22
List of workers can give wrong result


jirihnidek:
**Describe the bug**
List of workers can be incomplete or it can contain workers that are not active anymore

**To Reproduce**
1. Start yggdrasil on Terminal 1: `go run ./cmd/yggd --server tcp://localhost:1883 --log-level trace --client-id $(hostname)`
2. Start echo worker on Terminal 2: `go run ./worker/echo -log-level trace`
3. Try to get list of workers on Terminal 3 using: `go run ./cmd/yggctl workers list` or `busctl --user call com.redhat.Yggdrasil1 /com/redhat/Yggdrasil1 com.redhat.Yggdrasil1 ListWorkers`

**Current result**
List of workers is empty. The worker gets into the list of workers only after receiving the first message from MQTT.

**Expected behavior**
List of running workers.

Something like this in the case of `yggctl`:
```
echo - DispatchedAt:
echo - Version: 1
```

Something like this in the case of `busctl` (or any other D-Bus app):
```
a{sa{ss}} 1 "echo" 2 "Version" "1" "DispatchedAt" ""
```

**Additional Information**
 - OS: Fedora 38
 - Version: main branch

When worker has been stopped, then you can still see worker in the list of active workers despite MQTT message for the worker has been received and yggdrasil wasn't able to dispatch data from the MQTT message.

jirihnidek:
Started discussion that is related to this issue: https://github.com/RedHatInsights/yggdrasil/discussions/169

subpop:
Each worker is expected to own a name on the bus under the namespace `com.redhat.Yggdrasil1.Worker1` (i.e. `com.redhat.Yggdrasil1.Worker1.echo`). When asked for a list of workers, `yggd` should enumerate all names on the bus, filter for names that begin with the namespace value above, and return the results.

PR #252, opened 2024-08-22
Collect workers immediately

subpop:
The current implementation of `yggd` could be described as "lazy discovery of workers"; a worker is only discovered when a message arrives destined for it and the message is routed. If and only if `yggd` is able to successfully deliver a message to a worker, is it then added to the dispatchers table, triggering a publish of an updated "connection-status" message. This lazy-loading causes `yggctl` to output incorrect worker status under certain conditions. If `yggctl workers list` is run immediately after starting `yggd`, the list will be empty.

This PR changes the procedure `yggd` uses to detect and collect workers. After connecting to the bus (in the `(Dispatcher).Connect` method), a goroutine is run that scans the list of names on the bus, collecting any that begin with "com.redhat.Yggdrasil1.Worker1". After the workers are found, a "connection-status" message is published.

Additionally, a MatchSignal is added that watches for name ownership change (specifically the `org.freedesktop.DBus.NameOwnerChanged` signal) within the "com.redhat.Yggdrasil1.Worker1" namespace. If name ownership changes occur, the features are deleted or updated, causing a new "connection-status" message to publish.

I also discovered a related bug in the table and text output formats. If a worker had an empty "features" map, the "text" and "table" would erroneously omit those workers from the output because output was based on a for-loop over the features map. I chose to drop the "text" output format entirely, since it's functionally identical to "table". I also reformatted how the features map is included in the table output.

#### Old

```
WORKER  FIELD         VALUE
echo    DispatchedAt
echo    Version       1
```

#### New

```
WORKER  FEATURES
echo    {"DispatchedAt":"","Version":"1"}
```

Fixes: #135
Card ID: CCT-115

### DIFF
--- a/internal/work/dispatcher.go
+++ b/internal/work/dispatcher.go
@@ -130,9 +130,8 @@ func (d *Dispatcher) Connect() error {
 			case "org.freedesktop.DBus.Properties.PropertiesChanged":
 				changedProperties, ok := s.Body[1].(map[string]dbus.Variant)
 				if !ok {
-					log.Errorf(
-						"cannot convert body element 1 (changed_properties) to map[string]dbus.Variant: %v",
-						err,
+					log.Error(
+						"cannot convert body element 1 (changed_properties) to map[string]dbus.Variant",
 					)
 					continue
 				}
--- a/internal/work/dispatcher.go
+++ b/internal/work/dispatcher.go
@@ -113,6 +113,19 @@ func (d *Dispatcher) Connect() error {
 		)
 	}
 
+	err = d.conn.AddMatchSignal(
+		dbus.WithMatchObjectPath("/org/freedesktop/DBus"),
+		dbus.WithMatchInterface("org.freedesktop.DBus"),
+		dbus.WithMatchMember("NameOwnerChanged"),
+		dbus.WithMatchArg0Namespace("com.redhat.Yggdrasil1.Worker1"),
+	)
+	if err != nil {
+		return fmt.Errorf(
+			"cannot add signal match on org.freedesktop.DBus.NameOwnerChanged: %v",
+			err,
+		)
+	}
+
 	// start goroutine that receives values on the signals channel and handles
 	// them appropriately.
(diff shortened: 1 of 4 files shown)


internal/work/dispatcher.go
```diff
@@ -179,8 +186,89 @@ func (d *Dispatcher) Connect() error {
 						log.Errorf("cannot add journal entry: %v", err)
 					}
 				}()
+			case "org.freedesktop.DBus.NameOwnerChanged":
+				name, ok := s.Body[0].(string)
+				if !ok {
+					log.Errorf("cannot convert body element 0 (name) to string")
```
DuckBoss, line 192: These should be `log.Error` instead of `log.Errorf` similar to this commit e4811b7

internal/work/dispatcher.go
```diff
@@ -179,8 +186,89 @@ func (d *Dispatcher) Connect() error {
...
+					continue
+				}
+				oldOwner, ok := s.Body[1].(string)
+				if !ok {
+					log.Errorf("cannot convert body element 1 (old_owner) to string")
```
DuckBoss, line 197: Same nitpick here ^

internal/work/dispatcher.go
```diff
@@ -179,8 +186,89 @@ func (d *Dispatcher) Connect() error {
...
+					continue
+				}
+				newOwner, ok := s.Body[2].(string)
+				if !ok {
+					log.Errorf("cannot convert body element 2 (new_owner) to string")
```
DuckBoss, line 202: And here ^

DuckBoss (changes requested):
Thanks for the updates, this works great and it almost LGTM.
I have some small nitpicks before merging though.

internal/work/dispatcher.go
subpop: Good catch. Thank you.

DuckBoss (approved):
LGTM :+1:

jirihnidek (approved):
LGTM

PR #264, opened 2024-09-23
refactor: move sysusers file to dist/srpm

subpop:
Downstream distributions like Fedora prefer to manage sysuser files
downstream, so there is not much advantage to shipping a sysusers file
upstream. Instead move it to the in-tree 'dist/srpm' directory to
include the sysusers file when creating an RPM.

This commit also applies the Fedora %sysusers_create_compat macro to
create sysusers during installation.

Fixes: RHEL-59163


jirihnidek (approved):
LGTM it seems that it also fixes https://issues.redhat.com/browse/CCT-711

PR #272, opened 2024-10-25
fix: yggctl list workers should only show legacy names.

jvlcek:
Card ID" CCT-863

Hyphenated worker names were introduced to satisfy a console-dot assumption that worker names were hyphenated, after which the "yggctl workers list" command was showing both the "legacy" and hyphenated  worker names. Since hyphens in worker names are disallowed by the D-Bus naming policy it is safe to omit any worker names with hyphens when listing them.

### DIFF
--- a/cmd/yggctl/actions.go
+++ b/cmd/yggctl/actions.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"text/tabwriter"
 	"text/template"
 
@@ -183,6 +184,19 @@ func workersAction(c *cli.Context) error {
 	var workers map[string]map[string]string
 	if err := obj.Call("com.redhat.Yggdrasil1.ListWorkers", dbus.Flags(0)).Store(&workers); err != nil {
 		return cli.Exit(fmt.Errorf("cannot list workers: %v", err), 1)
+	}
+
+	// Remove worker names with hyphens.
+	// Why:
+	// Hyphenated worker names were introduced to satisfy a console-dot assumption that
+	// worker names were hyphenated, after which the "yggctl workers list" command was
+	// showing both the "legacy" and hyphenated  worker names. Since hyphens in worker
+	// names are disallowed by the D-Bus naming policy it is safe to remove any worker
+	// names with hyphens when listing them.
+	for key, _ := range workers {
+		if strings.ContainsRune(key, '-') {
+			delete(workers, key)
+		}
 	}
 
 	switch c.String("format") {


jvlcek:
@jirihnidek or @subpop Can one of you, or anyone else, please help me understand the lint failure? 

It expands to:

```
Run if [ "$(golines . --dry-run | wc -l)" -gt 0 ]; then exit 1; fi
  if [ "$(golines . --dry-run | wc -l)" -gt 0 ]; then exit 1; fi
  shell: /usr/bin/bash -e {0}
Error: Process completed with exit code 1.
```

It's not obvious to me what this is reporting on.

subpop:
> @jirihnidek or @subpop Can one of you, or anyone else, please help me understand the lint failure?
> 
> It expands to:
> 
> ```
> Run if [ "$(golines . --dry-run | wc -l)" -gt 0 ]; then exit 1; fi
>   if [ "$(golines . --dry-run | wc -l)" -gt 0 ]; then exit 1; fi
>   shell: /usr/bin/bash -e {0}
> Error: Process completed with exit code 1.
> ```
> 
> It's not obvious to me what this is reporting on.

The error is not useful, no. Line length is not enforced by `go fmt`, so [we run](https://github.com/RedHatInsights/yggdrasil/blob/2bbea96df07b9f27a96c3ea5cfff31e70962235c/CONTRIBUTING.md#code-guidelines) an additional linter ([`golines`](https://github.com/segmentio/golines)), to force wrap long lines. I suggest configuring your editor, IDE, or git pre-commit hook to run `go fmt` and `golines` before committing.

cmd/yggctl/actions.go
```diff
@@ -185,6 +186,12 @@ func workersAction(c *cli.Context) error {
 		return cli.Exit(fmt.Errorf("cannot list workers: %v", err), 1)
 	}
```
jirihnidek, line 188: I would add here some comment explaining, why we delete such workers. At the first glance, it looks strange that we do something like this. It has some historical reason and it should be mentioned here in comment.

jirihnidek (changes requested):
LGTM, I have only small request.

cmd/yggctl/actions.go
jvlcek: Thank you @jirihnidek and @subpop for the feedback.

I had come from the Ruby world where comments were highly frowned upon. But since I'm an old `C` programmer I surely can do comments! 😄 and I will update the commit message.

jvlcek: @jirihnidek and @subpop 

I believe this has been resolved by commit: https://github.com/RedHatInsights/yggdrasil/pull/272/commits/152af66a40e2c4f846fb69be4b37c139902acc23

Please take a look.

Thank you again for the help.

jvlcek:
@jirihnidek This is showing up as `Merging is blocked` due to requested changes but I think I have addressed all of them. Please let me know if I still need to address something else.

Thank you! Joe

jirihnidek (approved):
LGTM

jirihnidek:
@subpop is there anything else that blocks this PR from merging?

PR #277, opened 2024-11-07
build(rpm): Include local state dir

subpop:
Include the local state dir in the RPM so that the directory is created
when the package is installed. This enables other services that want to
write to the directory before the yggdrasil service starts.


jirihnidek (approved):
LGTM

PR #278, opened 2024-11-08
feat: Create facts file if not exists

Glutexo:
If the facts file does not exists upon start, it can’t be watched for changes even if created later. An empty facts dictionary is thus saved if non-existent path is provided.

Card IDs:
* [CCT-861](https://issues.redhat.com/browse/CCT-861)

### DIFF
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"crypto/tls"
+	"encoding/json"
 	"fmt"
 	"os"
 	"os/signal"
@@ -117,6 +118,35 @@ func setupClientID() error {
 		}
 		config.DefaultConfig.ClientID = string(clientID)
 	}
+	return nil
+}
+
+// setupFactsFile creates a canonical facts file if it doesn’t exist
+func setupFactsFile() error {
+	if config.DefaultConfig.FactsFile == "" {
+		return nil
+	}
+
+	_, err := os.Stat(config.DefaultConfig.FactsFile)
+	if err == nil {
+		return nil
+	}
+	if !os.IsNotExist(err) {
+		return fmt.Errorf("cannot stat file '%v': %v", config.DefaultConfig.FactsFile, err)
+	}
+
+	dir := filepath.Dir(config.DefaultConfig.FactsFile)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return fmt.Errorf("cannot create directory '%v': %v", dir, err)
+	}
+
+	facts := make(map[string]interface{})
+	data, _ := json.Marshal(facts)
+	if err := os.WriteFile(config.DefaultConfig.FactsFile, data, 0644); err != nil {
+		return fmt.Errorf("cannot create facts file '%v': %v", config.DefaultConfig.FactsFile, err)
+	}
+
+	log.Infof("facts file not found, created '%v'", config.DefaultConfig.FactsFile)
 	return nil
 }
 
@@ -356,6 +386,11 @@ func mainAction(c *cli.Context) error {
 	err = setupClientID()
 	if err != nil {
 		return err
+	}
+
+	err = setupFactsFile()
+	if err != nil {
+		return cli.Exit(fmt.Errorf("cannot setup facts file: %v", err), 1)
 	}
 
 	// Create HTTP client and TLS configuration. HTTP client is used for


Glutexo:
This is my first contribution to Yggdrasil. Because of that, I‘d like to ask for a review of the production code. There are no tests yet.

I wanted to keep the changeset small and relevant. This pull request thus does not fix the unmarshaling error in case of no facts file.

Glutexo:
I however think that `ConnectionStatus` is not the right place where the fact file should be created. There should be a separate function for that and if `ConnectionStatus` runs asynchronously, it should wait for that function to finish.

Glutexo:
Do I understand it correctly, that the `ConnectionStatus` function is untestable in its current state? It feels like instead of referencing `config.DefaultConfig.FactsFile`, a `config` variable would need to be passed as an argument.

Glutexo:
I would place the code to `mainAction`. There I’d go after `setupClientID()`, which also creates a file, and before `setupTLS()`, the first step towards creating the HTTP client.

Glutexo:
I see there are no tests for `main.go`. If my new function `setupFactsFile` were to be tested, it would require passing `config.DefaultConfig` as an argument. Similar functions like `setupClientID` would need the same. That means making my function testable would introduce inconsistency.

Hence, I would omit the tests in this pull request. I’d rather create a ticket to make the module testable and write the tests. **This pull request is thus ready for review.**

cmd/yggd/main.go
```diff
@@ -120,6 +121,32 @@ func setupClientID() error {
...
+	}
+
+	_, err := os.Stat(config.DefaultConfig.FactsFile)
+	if !os.IsNotExist(err) {
+		return err
+	}
+
+	dir := filepath.Dir(config.DefaultConfig.FactsFile)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return err
```
subpop, line 137: Similarly,

```suggestion
		return fmt.Errorf("cannot create directory '%v': %v", dir, err)
```

cmd/yggd/main.go
```diff
@@ -120,6 +121,32 @@ func setupClientID() error {
...
+	}
+
+	facts := make(map[string]interface{})
+	data, _ := json.Marshal(facts)
+	if err := os.WriteFile(config.DefaultConfig.FactsFile, data, 0644); err != nil {
+		return err
```
subpop, line 143:
```suggestion
		return fmt.Errorf("cannot create facts file '%v': %v", config.DefaultConfig.FactsFile, err)
```

cmd/yggd/main.go
```diff
@@ -358,6 +385,12 @@ func mainAction(c *cli.Context) error {
 		return err
 	}
 
+	fmt.Println("***")
```
subpop, line 388: This looks like a marker line that should not be committed?

cmd/yggd/main.go
```diff
@@ -358,6 +385,12 @@ func mainAction(c *cli.Context) error {
...
+	err = setupFactsFile()
+	if err != nil {
+		return err
```
subpop, line 391: Same thing wrapping the error message here.

subpop (changes requested):
Looks really good. You're right about the testability of the code. It's not very well covered by unit tests, and writing them was not feasible given the timeline pressure the project has historically been under. Creating a follow up issue (likely an epic) to refactor the code the be more testable is the right way to approach this.

I left a few comments about the commit, but overall, this is probably ready to merge soon.

cmd/yggd/main.go
Glutexo: Oh.

cmd/yggd/main.go
Glutexo: I was not sure whether the error message should be wrapped here. Some calls from `mainAction` do,

```go
	client, transporter, err := setupClient(dispatcher, tlsConfig)
	if err != nil { return cli.Exit(fmt.Errorf("cannot setup client: %w", err), 1) }
```

some don’t.


```go
	err = setupClientID()
	if err != nil { return err }
```

I think it’s good practice to wrap and there are already wrapped ones. I put the wrapper there, then. It allows to add a single `cli.Exit()` wrapper too.

I’ll create a card for consolidating the error returns.

cmd/yggd/main.go
```diff
@@ -120,6 +121,32 @@ func setupClientID() error {
...
 
+// setupFactsFile creates a canonical facts file if it doesn’t exist
+func setupFactsFile() error {
+	if config.DefaultConfig.FactsFile == "" {
+		return nil
+	}
+
+	_, err := os.Stat(config.DefaultConfig.FactsFile)
+	if !os.IsNotExist(err) {
+		return err
```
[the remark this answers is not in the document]
Glutexo: The conditional here catches both `nil` (file exists) and `!os.IsNotExist(err)` (a problem). The wrapper is only appropriate for the real error, which is the latter case. Thus I had to split this conditional.

Glutexo:
Thanks for your review, @subpop! I fixed the issues you spotted.

I also realized one thing: if the facts file exists, but can’t be read, my `setupFactsFile` function doesn’t fail. `os.Stat` only reads metadata of the file, not its contents. That means read permissions are not required.

The missing permissions will however make watcher setup fail later.

```
yggdrasil/cmd/yggd/client.go:388: cannot read facts file: open facts-unreadable.json: permission denied
```

This was the case before my changes already. I didn’t change the behavior by trying to open the file for reading. But I think it’s a good idea to do so: I’d rather create a separate card instead.

Glutexo:
The failing check appears not to be related to this PR. It’a VM provisioning failure.

cmd/yggd/main.go
```diff
@@ -120,6 +121,35 @@ func setupClientID() error {
...
+		return fmt.Errorf("cannot create directory '%v': %v", dir, err)
+	}
+
+	facts := make(map[string]interface{})
+	data, _ := json.Marshal(facts)
+	if err := os.WriteFile(config.DefaultConfig.FactsFile, data, 0644); err != nil {
+		return fmt.Errorf("cannot create facts file '%v': %v", config.DefaultConfig.FactsFile, err)
+	}
+
+	log.Infof("facts file not found, created")
```
subpop, line 149: Hey, something I just thought of that might be useful here. Can we print the file name as part of this info message?
```suggestion
	log.Infof("facts file not found, created %v", config.DefaultConfig.FactsFile)
```

Glutexo: Great idea!

PR #281, opened 2024-11-22
fix: adjust release to include git info

subpop:
Adjust the forge macros so they work across EL9, 10 and Fedora.


jirihnidek (approved):
LGTM

PR #283, opened 2024-12-06
Add automatic release workflows

subpop:
Add two workflows:
- update-version
- tag-release

`update-version` triggers monthly on the first day of the month, or can be triggered manually. It checks out the repository, calculates a version incrementing the "z" value, rewrite's meson.build with the new value, commits the file to a branch and opens a PR submitting the commit. It sets the label auto-release on the PR.

`tag-release` triggers on pull requests, and only if the PR is merged and contains a label auto-release. If this workflow runs, it configures the project, generates a distribution tarball and creates a release, attaching the distribution tarball to the release.

This PR also deletes the dist.yml workflow, as tag-release.yml replaces it.

Card ID: [CCT-1012](https://issues.redhat.com/browse/CCT-1012)


jirihnidek (approved):
LGTM

PR #288, opened 2025-01-15
List Activatable Workers on startup

subpop:
When the Dispatcher connects, find all activatable workers instead of only active/running workers. This will include workers that are installed but that have not yet started.

Card ID: CCT-862

### DIFF
--- a/internal/work/dispatcher.go
+++ b/internal/work/dispatcher.go
@@ -243,7 +243,7 @@ func (d *Dispatcher) Connect() error {
 	// start goroutine that finds workers already active on the bus connection
 	// and get their features.
 	go func() {
-		workers, err := d.findWorkers()
+		workers, err := d.findActivatableWorkers()
 		if err != nil {
 			log.Errorf("cannot find workers: %v", err)
 			return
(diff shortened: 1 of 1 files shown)


jirihnidek:
@subpop When I start `yggdrasil.service` from this PR, then all installed yggdrasil workers are automatically started. This could have unintentional side-effect of high consumption of memory and other resources. I believe that worker should be started only in situation, when it is needed.

Archana-PandeyM:
@subpop In CCt-883 : Acceptance Criteria:
Immediately after starting yggdrasil.service, all workers, regardless of their running state, are discovered and included as part of its "dispatchers" map. These workers are included in the published "connection-status" message.

If we decide not to go by the acceptance criteria then the rhc experience will change- i.e. we need to document that workers need to be started outside of yggdrasil. 

or we could make an exception for rhc-worker-playbook and yggdrasil-worker-package-manager

jirihnidek:
Not merged to main branch, because we are waiting for approval from QE team.

Archana-PandeyM:
LGTM.

PR #294, opened 2025-01-30
feat: Support an empty config file

jvlcek:
**Card:** CCT-648

### SUMMARY:

This PR supports delivering yggd with fully commented out config file.
and
The hardcoded, default, broker protocol is being changed from mqtt to none.

### CHANGE DETAILS

**_See file(s):_**

-   modified:   internal/config/config.go

A small comment change was made to include the supported values for the broker protocol.

**_See file(s):_**

-   modified:   CONTRIBUTING.md
-   modified:   README.md

The CONTRIBUTING and README pages were updated to reflect the functional changes.

**_See file(s):_**

-   modified:   cmd/yggd/main.go

The hardcoded, default, broker protocol is changed from mqtt to none.

**_See file(s):_**

-   modified: data/yggdrasil/config.toml

The file /etc/yggdrasil/config.toml now contains all commented out lines.

**_See file(s):_**

-   modified:   dist/srpm/yggdrasil.spec.in

This file is updated to specifically create the /etc/yggdrasil directory.

### TESTING:

To test this PR use the RPM from the automated CI/CD build.

**Test 1:** Read the CONTRIBUTING.md and README.md file for clarity and readability.

**Test 2:** Start the yggdrasil service and ensure the new hardcoded default protocol of none is used.

Install the RPM from the automated CI/CD build and then:

Start the yggdrasil service and ensure the new hardcoded default protocol of none is used.

**Test 3:** Confirm /etc/yggdrasil/config.toml is created with all lines commented out

Confirm /etc/yggdrasil/config.toml exists and all lines are commented out.
e.g.:

```
cat /etc/yggdrasil/config.toml
# yggdrasil global configuration settings
#
# protocol = "mqtt"
# server = ["tcp://test.mosquitto.org:1883"]
# log-level = "error"
```

**Test 4:** Confirm /etc/yggdrasil/config.toml is saved to /etc/yggdrasil/config.toml.rpmsave
        when the RPM is uninstalled.

uninstall the yggdrasil RPM and confirm the config.toml file was saved as /etc/yggdrasil/config.toml.rpmsave

```
cat /etc/yggdrasil/config.toml.rpmsave
# yggdrasil global configuration settings
#
# protocol = "mqtt"
# server = ["tcp://test.mosquitto.org:1883"]
# log-level = "error"
```

**Test 5:** Confirm /etc/yggdrasil/config.toml is preserved if it has non-commented out content.

uninstall the yggdrasil RPM

remove all files in directory /etc/yggdrasil/

Create the file /etc/yggdrasil/config.toml with one, uncommented, line.

e.g.:

```
$ cat /etc/yggdrasil/config.toml*
Bob = "Your Uncle"
```

Install the RPM from the automated CI/CD build and then:

Confirm /etc/yggdrasil/config.toml is unchanged and file /etc/yggdrasil/config.toml.rpmnew
was created.

e.g.:

```
$ cat /etc/yggdrasil/config.toml
Bob = "Your Uncle"

$ cat /etc/yggdrasil/config.toml.rpmnew
# yggdrasil global configuration settings
#
# protocol = "mqtt"
# server = ["tcp://test.mosquitto.org:1883"]
# log-level = "error"
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,8 +111,14 @@ sudo go run ./cmd/yggd --server tcp://localhost:1883 --log-level trace --client-
 
 `yggd` can be compiled using meson, or can be run directly with the `go run`
 command. It can read configuration values from a file by running
-`yggd` with the `--config` option. A sample configuration file is included in
-the `data/yggdrasil` directory.
+`yggd` with the `--config` option. If this flag is not specified
+by default the file location, `/etc/yggdrasil/config.toml`, is used.
+A sample configuration file, with all of the lines commented out,
+is included in the the source tree under the `data/yggdrasil`
+directory. This file is installed in the default file location,
+`/etc/yggdrasil/config.toml`.
+To use this sample configuration file uncomment the desired lines and
+run it with the following command:
 
 ```
 sudo go run ./cmd/yggd --config ./data/yggdrasil/config.toml
@@ -144,7 +150,9 @@ port using firewalld, run:
 sudo firewall-cmd --zone public --add-port 2345/tcp --permanent
 ```
 
-Start `dlv` using the `debug` command:
+Start `dlv` using the `debug` command, with the desired lines in
+the sample configuration file commented out, then run it with
+the following command:
 
 ```
 sudo /root/go/bin/dlv debug --api-version 2 --headless --listen 0.0.0.0:2345 \
--- a/README.md
+++ b/README.md
@@ -36,9 +36,11 @@ Configuration of `yggd` can be done by specifying values in a configuration file
 or via command line arguments. Command-line arguments take precedence over
 configuration file values. The configuration file is [TOML](https://toml.io).
 
-The system-wide configuration file is located at `/etc/yggdrasil/config.toml`
-(assuming `SYSCONFDIR=/etc`, as the example above). The location of the file may
-be overridden by passing the `--config` command-line argument to `yggd`.
+The default location searched for the system-wide configuration file is at
+`/etc/yggdrasil/config.toml` (assuming SYSCONFDIR=/etc, as the example above).
+
+The default location of the file may be overridden by passing the `--config`
+command-line argument to `yggd`.
 
 ### (Optional) Authentication
 
@@ -73,6 +75,8 @@ when connecting to the broker. To do this, create the file
 The contents of the file may be any number of TOML key/value pairs. However, a
 limited number of TOML values are accepted as tag values (strings, integers,
 booleans, floats, Local Date, Local Time, Offset Date-Time and Local Date-Time).
+An example the tags.toml file can be found at at
+/usr/share/doc/yggdrasil/tags.toml
 
 ## Running
 
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -524,7 +524,7 @@ func main() {
 		altsrc.NewStringFlag(&cli.StringFlag{
 			Name:  config.FlagNameProtocol,
 			Usage: "Transmit data remotely using `PROTOCOL` ('mqtt', 'http' or 'none')",
-			Value: "mqtt",
+			Value: "none",
 		}),
 		altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
 			Name:  config.FlagNameServer,
--- a/data/yggdrasil/config.toml
+++ b/data/yggdrasil/config.toml
@@ -1,5 +1,5 @@
 # yggdrasil global configuration settings
-
-protocol = "mqtt"
-server = ["tcp://test.mosquitto.org:1883"]
-log-level = "error"
+#
+# protocol = "mqtt"
+# server = ["tcp://test.mosquitto.org:1883"]
+# log-level = "error"
--- a/dist/srpm/yggdrasil.spec.in
+++ b/dist/srpm/yggdrasil.spec.in
@@ -107,6 +107,7 @@ export %gomodulesmode
 %meson_install
 install -p -D -m 0644 %{SOURCE1} %{buildroot}%{_sysusersdir}/%{name}.conf
 install -d -m 0755 %{buildroot}%{_localstatedir}/lib/yggdrasil
+install -d -m 0755 %{buildroot}%{_sysconfdir}/yggdrasil
 %if %{has_go_rpm_macros}
 %gopkginstall
 %endif
@@ -148,6 +149,7 @@ install -d -m 0755 %{buildroot}%{_localstatedir}/lib/yggdrasil
 %{_datadir}/dbus-1/{interfaces,system-services,system.d}/*
 %{_datadir}/doc/%{name}/*
 %{_mandir}/man1/*
+%{_sysconfdir}/yggdrasil
 %attr(0755, yggdrasil, yggdrasil) %{_localstatedir}/lib/yggdrasil
 
 %if %{has_go_rpm_macros}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -65,7 +65,7 @@ type Config struct {
 	PathPrefix string
 
 	// Protocol is the protocol used by yggd when connecting to Server. Can be
-	// either MQTT or HTTP.
+	// either MQTT, HTTP or none.
 	Protocol string
 
 	// DataHost is a hostname value to interject into all HTTP requests when


jirihnidek:
I did some experiment with installing generated RPM. When the yggdrasil was first installed from Fedora repo and the `/etc/yggdrasil/config.toml` and then you install generated RPM, then old `/etc/yggdrasil/config.toml` file is deleted as you can see:

```
user@localhost:~/$ sudo dnf install /var/lib/mock/fedora-41-x86_64/result/yggdrasil-0.4.4-99.20250214git82b5ac0.fc41.x86_64.rpm
Updating and loading repositories:
Repositories loaded.
Package                                                          Arch            Version                                                           Repository                                Size
Upgrading:
 yggdrasil                                                       x86_64          0.4.4-99.20250214git82b5ac0.fc41                                  @commandline                          20.5 MiB
   replacing yggdrasil                                           x86_64          0.4.4-3.fc41                                                      fedora                                16.5 MiB

Transaction Summary:
 Upgrading:          1 package
 Replacing:          1 package

Total size of inbound packages is 6 MiB. Need to download 0 B.
After this operation, 4 MiB extra will be used (install 21 MiB, remove 16 MiB).
Is this ok [y/N]: y
Running transaction
[1/4] Verify package files                                                                                                                               100% |  23.0   B/s |   1.0   B |  00m00s
[2/4] Prepare transaction                                                                                                                                100% |   2.0   B/s |   2.0   B |  00m01s
[3/4] Upgrading yggdrasil-0:0.4.4-99.20250214git82b5ac0.fc41.x86_64                                                                                      100% | 121.5 MiB/s |  20.5 MiB |  00m00s
[4/4] Removing yggdrasil-0:0.4.4-3.fc41.x86_64                                                                                                           100% |  12.0   B/s |  30.0   B |  00m02s
>>> [RPM] /etc/yggdrasil/config.toml saved as /etc/yggdrasil/config.toml.rpmsave
Warning: skipped OpenPGP checks for 1 package from repository: @commandline
Complete!

user@localhost:~/$ cat /etc/yggdrasil/config.toml
cat: /etc/yggdrasil/config.toml: No such file or directory
```

I think that existing `/etc/yggdrasil/config.toml` should not be deleted during update of yggdrasil RPM (despite it is saved as `/etc/yggdrasil/config.toml.rpmsave`. This file should just persist.

jirihnidek (changes requested):
Thanks for the PR. :+1: Overall it does what it should do. I have some comments and suggestions to provide better user experience. These questions and suggestions are more for @subpop .

It is not easy to find the example of config file. The `/usr/share/doc/yggdrasil/example_mqtt_config.toml` would not be probably the first place I would look. Wouldn't be better approach to create file `/etc/yggdrasil/config.toml` with uncommented configuration options. Something like this:

```toml
# yggdrasil global configuration settings
#
# protocol = "mqtt"
# server = ["tcp://test.mosquitto.org:1883"]
# log-level = "error"
```

It would be easy to find and easy to edit and use. Maybe, we could add there more options, because `yggd` supports more options:

```
root@localhost:/# yggd --help
NAME:
   yggd - connect the system to the network in order to send and receive messages

USAGE:
   yggd [global options] command [command options] 

VERSION:
   0.4.4

COMMANDS:
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --config FILE                  Read config values from FILE
   --log-level LEVEL              Set the logging output level to LEVEL (default: "info")
   --cert-file FILE               Use FILE as the client certificate
   --key-file FILE                Use FILE as the client's private key
   --protocol PROTOCOL            Transmit data remotely using PROTOCOL ('mqtt', 'http' or 'none') (default: "none")
   --server URI [ --server URI ]  Connect the client to the specified URI
   --data-host HOST               Force all HTTP traffic over HOST
   --client-id VALUE              Use VALUE as the client ID when connecting
   --facts-file FILE              Read facts from FILE
   --message-journal FILE         Record worker events and messages in the database FILE
   --help, -h                     show help
   --version, -v                  print the version
```

Or even better, could we autogenerate this uncommented configuration file with comments explaining each option?

subpop:
> It is not easy to find the example of config file. The `/usr/share/doc/yggdrasil/example_mqtt_config.toml` would not be probably the first place I would look. Wouldn't be better approach to create file `/etc/yggdrasil/config.toml` with uncommented configuration options. Something like this:
> 
> ```toml
> # yggdrasil global configuration settings
> #
> # protocol = "mqtt"
> # server = ["tcp://test.mosquitto.org:1883"]
> # log-level = "error"
> ```
> Or even better, could we autogenerate this uncommented configuration file with comments explaining each option?

I like both of these ideas. To keep the scope of this PR low, let's manually install a commented out config file in `/etc/yggdrasil/config.toml`. The idea of auto-generating a file is interesting, but might require more work to get right. (We might actually need to add that functionality to `urfave/cli` first in order to then use it downstream here.)

jvlcek:
> > It is not easy to find the example of config file. The `/usr/share/doc/yggdrasil/example_mqtt_config.toml` would not be probably the first place I would look. Wouldn't be better approach to create file `/etc/yggdrasil/config.toml` with uncommented configuration options. Something like this:
> > ```toml
> > # yggdrasil global configuration settings
> > #
> > # protocol = "mqtt"
> > # server = ["tcp://test.mosquitto.org:1883"]
> > # log-level = "error"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or even better, could we autogenerate this uncommented configuration file with comments explaining each option?
> 
> I like both of these ideas. To keep the scope of this PR low, let's manually install a commented out config file in `/etc/yggdrasil/config.toml`. The idea of auto-generating a file is interesting, but might require more work to get right. (We might actually need to add that functionality to `urfave/cli` first in order to then use it downstream here.)

I believe the issue with doing this will complicate what happens when updating the RPM.

Today if a config.toml or tags.toml file exists in the /etc/yggdrasil/ directory installing the newer RPM will result in these files being copied to the .rpmsave file name and the files built into the RPM will replace them. As this PR stands now that sort of still happens. It's just that there is not a new version of the the config.toml.

So my question is:

Should we keep the existing behavior of copying any existing config.toml or tags.toml to config.toml.rpmsave or tags.toml.rpmsave? 

Or should we not copy in the new template file?

Thoughts?

subpop:
> So my question is:
> 
> Should we keep the existing behavior of copying any existing config.toml or tags.toml to config.toml.rpmsave or tags.toml.rpmsave?
> 
> Or should we not copy in the new template file?
> 
> Thoughts?

The .rpmsave file is created automatically by the `rpm` program during an upgrade. [This chapter of the Max RPM document](http://ftp.rpm.org/max-rpm/ch-rpm-upgrade.html) describes it best. In our case, I think it's safe to mark the file as `%config` and let RPM figure out the right way to handle the file during an upgrade.

jvlcek:
Thank you for the feedback @jirihnidek and @subpop 

I've made the following updates:

 063bf99  -  Return to delivering sample data/yggdrasil/config.toml but commented out
 b183f5d  -  Remove the reference to example_mqtt_config.toml
 89f4042  -  Return the config.toml exampl to the data directory.

I've also updated the top comment to describe the changes to each updated file. I also added a section describing the testing I have done.


I believe these changes address your feedback.

Please take a look.

Thank you for the help!

Joe

jvlcek:
> > So my question is:
> > Should we keep the existing behavior of copying any existing config.toml or tags.toml to config.toml.rpmsave or tags.toml.rpmsave?
> > Or should we not copy in the new template file?
> > Thoughts?
> 
> The .rpmsave file is created automatically by the `rpm` program during an upgrade. [This chapter of the Max RPM document](http://ftp.rpm.org/max-rpm/ch-rpm-upgrade.html) describes it best. In our case, I think it's safe to mark the file as `%config` and let RPM figure out the right way to handle the file during an upgrade.

Thank you @subpop 

My question was based on my lack of correct understanding of how config files are managed by updating RPMs. I believe the current implementation does the correct thing in this regard.

jirihnidek (approved):
Thanks for updates. LGTM

jvlcek:
Thank you for the help @jirihnidek and @subpop 

Can someone please merge this for me? I don't have the proper authority to merge it.

jirihnidek:
@jvlcek we are waiting for approval from QE. I did some testing. It works from my point of view, but @Archana-PandeyM or @Lorquas or somebody else form QE team has to approve it.

subpop (approved):
This looks good to me as well. I would also like to test the condition where the config file is actually deleted too.

Archana-PandeyM:
On a connected system where yggdrasil service is active, upgrading the yggdrasil package replace the existing config.toml with the one that is present in upgraded rpm, Shouldn't we existing rpm restored?

subpop:
> On a connected system where yggdrasil service is active, upgrading the yggdrasil package replace the existing config.toml with the one that is present in upgraded rpm, Shouldn't the existing rpm restored? Current behavior breaks remediations and also the pre upgrade config file completely lost.

Oh hm. That might be because the config file was never customized. So the upgrade algorithm decided to replace it with the RPM-packaged one.

subpop:
Wait, Archana, are you upgrading from the RHEL-packaged RPM to the COPR built one? Those two use different config files. In the case of upgrading from a RHEL package to the new RHEL package, we can let this config file replacement happen by use compile-time defaults to ensure the behavior doesn't change. This means that we'll need to ensure that each config file option that we ship in the default config file in dist-git will need compile-time options to set their value correctly.

Archana-PandeyM:
Yes I was testing copr build on RHEL10. If this change is going to stay in copr only then lets merge it.

PR #300, opened 2025-02-07
fix: Do not print scary messages to log for old directive

jirihnidek:
* When MQTT message with old directive standard is received, then do not print scary message, but only print informative message that name was transformed
* Added one more unit test

### DIFF
--- a/internal/work/dbus.go
+++ b/internal/work/dbus.go
@@ -22,9 +22,8 @@ func NewDBusError(name string, body ...string) *dbus.Error {
 func ScrubName(name string) (string, error) {
 	r := regexp.MustCompile("-")
 	if r.Match([]byte(name)) {
-		return string(
-			r.ReplaceAll([]byte(name), []byte("_")),
-		), fmt.Errorf("invalid worker name: %v", name)
+		newName := string(r.ReplaceAll([]byte(name), []byte("_")))
+		return newName, fmt.Errorf("directive name '%v' transformed to '%v'", name, newName)
 	}
 	return name, nil
 }


internal/work/dbus_test.go
```diff
@@ -43,6 +43,12 @@ func TestScrubName(t *testing.T) {
 			want:        "alpha_bravo_charlie",
 			wantError:   cmpopts.AnyError,
 		},
+		{
+			description: "old rhc-worker-playbook",
+			input:       "rhc-worker-playbook",
+			want:        "rhc_worker_playbook",
+			wantError:   cmpopts.AnyError,
+		},
```
subpop, line 51: How is this test case any different from the other cases, specifically "multiple hyphens"?

internal/work/dbus.go
```diff
@@ -22,9 +22,10 @@ func NewDBusError(name string, body ...string) *dbus.Error {
 func ScrubName(name string) (string, error) {
 	r := regexp.MustCompile("-")
 	if r.Match([]byte(name)) {
+		newName := r.ReplaceAll([]byte(name), []byte("_"))
 		return string(
-			r.ReplaceAll([]byte(name), []byte("_")),
-		), fmt.Errorf("invalid worker name: %v", name)
+			newName,
+		), fmt.Errorf("worker name: %v transformed to %v", name, newName)
```
subpop, line 28: Given the rewritten error message, we should probably drop the colon:
```suggestion
		), fmt.Errorf("worker name %v transformed to %v", name, newName)
```

subpop (changes requested):

internal/work/dbus_test.go
jirihnidek: Yeah, I added this only for being sure that it works as expected for this worker. Removed this change in the latest version of this PR.

jirihnidek:
You can test this change when you run yggdrasil according `CONTRIBUTING.md`, but you try to send message with directive containing `-`. Such worker needn't to exists.

```
echo '"hello"' | go run ./cmd/yggctl generate data-message --directive foo-fighter - \
    | pub -broker tcp://localhost:1883 -topic yggdrasil/$(hostname)/data/in
```

Then you should see following log message in the output of yggdrasil:

```
[yggd] 2025/02/11 01:23:45 /path/to/yggdrasil/internal/work/dispatcher.go:304: \directive name 'foo-fighter' transformed to 'foo_fighter'
```

jirihnidek:
> Did this PR originate from a CCT issue?

I don't know.

@Archana-PandeyM did you create any card for this issue?

subpop (approved):

PR #309, opened 2025-03-05
Add dependabot config

subpop:


ptoscano:
From what I know and I see, dependabot configurations on non-default branches are simply ignored:
https://github.com/dependabot/dependabot-core/issues/2159
The "workaround" mentioned is to use more configurations in the default branch for different branches.

PR #316, opened 2025-03-13
build: set go version to 1.22.0

subpop:
Add a toolchain directive

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/redhatinsights/yggdrasil
 
-go 1.22
+go 1.22.0
+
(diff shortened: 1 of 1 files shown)


subpop:
I figured out why the rawhide builds were failing. It turns out `golang(github.com/docker/docker)` was removed from rawhide. `golang(github.com/golang-migrate/migrate)` required docker for testing. I updated the `golang-migrate` RPM in Fedora to remove the test files that were pulling in the docker requirement. Once this new version of `migrate` is in rawhide, these rawhide builds should start passing again.

PR #317, opened 2025-03-14
CI: Install dependencies and print go version

jirihnidek:
* Added two another steps to go GitHub action
  * Install dependencies
  * Print go version used for build


.github/workflows/go.yml
```diff
@@ -14,6 +14,10 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version: "stable"
+      - run: go version
+        with:
+          go-version: "stable"
```
subpop, line 19: Why did you specify a `with.go-version` for this step, but not the `go get .` step? What is the default go version used if you don't specify `with.go-version`? (My guess would be "stable".) Do you feel that this explicit `with.go-version` is necessary? Maybe we should add it to each step?

subpop (changes requested):

.github/workflows/go.yml
jirihnidek: We can drop `with.go-version` here, because documentation says:

```
To use a preinstalled version of Go on a GitHub-hosted runner, pass the relevant version to the go-version property of the setup-go action. This action finds a specific version of Go from the tools cache on each runner, and adds the necessary binaries to PATH. These changes will persist for the remainder of the job.
```

subpop (approved):

PR #323, opened 2025-04-01
fix: Avoid returning error when stopping a worker

jvlcek:
Card: CCT-1066


jvlcek:
When stopWorker detects an issue do not raise an error. Simply log a debug level message and return.

This will resolve issues Craig Donnelly uncovered during testing

Aftermath (approved):
LGTM. Tested against a RHEL 9.7 compose, and covered a current release broken scenario.
This moves some logs to debug-only output, in order to prevent alarming users to things that are not true errors, and allows us to complete cleanup of stale PIDs.

jirihnidek (approved):
LGTM

PR #325, opened 2025-04-02
fix: Added toolchain directive to go.mod

jirihnidek:
* The toolchain directive should exist in go.mod for go 1.21 and higher. Otherwise, some tools may not work properly


jirihnidek:
Closing, because @subpop had almost the same idea in #316

PR #333, opened 2025-04-30
Potential fix for code scanning alert no. 3: Workflow does not contain permissions

jirihnidek:
Potential fix for [https://github.com/RedHatInsights/yggdrasil/security/code-scanning/3](https://github.com/RedHatInsights/yggdrasil/security/code-scanning/3)

To fix the issue, we will add a `permissions` block at the root level of the workflow. This block will apply to all jobs in the workflow unless overridden by job-specific `permissions`. Since the workflow is for linting, it only requires read access to the repository contents. We will set `contents: read` as the minimal required permission.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   lint-commits:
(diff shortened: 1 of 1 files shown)


DuckBoss (approved):
LGTM

PR #344, opened 2025-06-25
feat: infer protocol from first server entry when unset

jlocash:
* Card ID: CCT-1448
* When the protocol flag is unset, i.e in /etc/yggdrasil/config.toml, it will be inferred from the first server entry (if set)

### DIFF
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -355,6 +356,27 @@ func systemdWatchDog() {
 	}
 }
 
+// detectProtocolFromURL attempts to detect and return the first valid protocol and URL from the list of server URLs
+func detectProtocolFromURL(serverURLs []string) (string, string, error) {
+	for _, serverURL := range serverURLs {
+		parsedURL, err := url.Parse(serverURL)
+		if err != nil {
+			log.Errorf("error parsing server URL '%s': %v", serverURL, err)
+			continue
+		}
+
+		switch parsedURL.Scheme {
+		case "http", "https":
+			return "http", serverURL, nil
+		case "mqtt", "mqtts":
+			return "mqtt", serverURL, nil
+		default:
+			log.Warnf("unsupported protocol '%s' in server URL '%s'", parsedURL.Scheme, serverURL)
+		}
+	}
+	return "", "", fmt.Errorf("no protocols detected from %q", serverURLs)
+}
+
 // mainAction is main action of yggd
 func mainAction(c *cli.Context) error {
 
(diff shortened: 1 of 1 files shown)


cmd/yggd/main.go
```diff
@@ -382,6 +383,18 @@ func mainAction(c *cli.Context) error {
 	}
 	log.Infof("starting %v version %v", c.App.Name, c.App.Version)
 
+	// When no protocol is defined in the config, infer it from the first server entry
+	if config.DefaultConfig.Protocol == "none" && len(config.DefaultConfig.Server) > 0 {
+		log.Warn("no network protocol specified - inferring from first server entry")
```
jirihnidek, line 388: I would alter this log message a little to something like:

```go
log.Warn("non network protocol specified in the '%s' file nor CLI option. Trying to parse protocol from server URL '%s'.",
c.String("config"),
config.DefaultConfig.Server[0])
```

Note: The `config.DefaultConfig.Server` is a slice of strings, because MQTT allows to specify more brokers. We should check consistency of URLs and specified protocol (if set in the `beforeAction()`), but we should do this in another PR.

BTW: The "infer" is not so common English word. A non-native English speakers may not know this world. I didn't know it.

cmd/yggd/main.go
```diff
@@ -382,6 +383,18 @@ func mainAction(c *cli.Context) error {
...
+		serverURL, err := url.Parse(config.DefaultConfig.Server[0])
+		if err != nil {
+			log.Errorf("error parsing server url '%s': %v", config.DefaultConfig.Server[0], err)
```
jirihnidek, line 391: Nitpick: url -> URL

cmd/yggd/main.go
```diff
@@ -382,6 +383,18 @@ func mainAction(c *cli.Context) error {
...
+		} else {
+			config.DefaultConfig.Protocol = serverURL.Scheme
+			log.Infof("inferred protocol '%s' from server '%s'", config.DefaultConfig.Protocol, config.DefaultConfig.Server[0])
```
jirihnidek, line 394: Nitpick: from server '%s' -> from server URL '%s'

jirihnidek (changes requested):
Thanks for PR. I have few comments and requests.

cmd/yggd/main.go
jlocash: How about something like:

```go
log.Warnf(
	"network protocol not specified in config file '%s' or CLI - detecting from server URL '%s'",
	c.String("config"),
	config.DefaultConfig.Server[0],
)
```

cmd/yggd/main.go
```diff
@@ -157,7 +158,7 @@ func setupClient(
 ) (*Client, transport.Transporter, error) {
 	var transporter transport.Transporter
 	switch config.DefaultConfig.Protocol {
-	case "mqtt":
+	case "mqtt", "mqtts":
```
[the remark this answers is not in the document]
jlocash: After thinking about this more, a user user could  specify `https` or `mqtts` as the protocol in their config, use no cert file, and use an `http://...` or `mqtt://...` server URL, and yggdrasil will still try to connect. Alternatively, I could keep the original cases as they were and add some additional checks when parsing the server URL:

```golang
...
if serverURL.Scheme == "mqtts" {
	config.DefaultConfig.Protocol = "mqtt"
}
if serverURL.Scheme == "https" {
	config.DefaultConfig.Protocol = "http"
}
...
```

jirihnidek (changes requested):
Thanks for the update. :+1:  I have another proposal how to improve the code.

cmd/yggd/main.go
```diff
@@ -382,6 +383,34 @@ func mainAction(c *cli.Context) error {
...
+		}
+
+		if config.DefaultConfig.Protocol != "none" {
+			log.Infof(
+				"detected protocol '%s' from server URL '%s'",
+				config.DefaultConfig.Protocol,
+				serverURL,
+			)
+		}
+	}
```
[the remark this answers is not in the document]
jirihnidek: It is OK now. Later we can add some validation of provided URLs. E.g. following cases would not be valid:

1. `http://foo.com https://bar.org`
2. `https://foo.com mqtt://foo.com`
3. `mqtt://foo.com https://bar.org`

But we can do it later in another PR.

Archana-PandeyM:
LGTM!

pondrejk (approved):
I ran the satellite client testing scenario using the packit built, yggdrasil got configured and started correctly on the client and the remote job executed as expected

PR #346, opened 2025-07-11
build: update go version to 1.24 in go.mod

DuckBoss:
* Card ID: RHEL-100953
* Update go version to 1.24 for default post-quantum X25519MLKEM768 key exchange.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/redhatinsights/yggdrasil
 
-go 1.23.0
+go 1.24.0
 
-toolchain go1.23.6
+toolchain go1.24.4
 
 require (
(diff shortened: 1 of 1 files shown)


jlocash (approved):
Changes LGTM

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/redhatinsights/yggdrasil
 
-go 1.23.0
+go 1.24.0
```
jirihnidek, line 3: Was this change really necessary? If I understand this correctly, then this line says that `yggdrasil` source code requires (directly or indirectly) go 1.24 to be able to compile it. Does really `yggdrasil` or required package require go 1.24? Isn't change of `toolchain` enough?

PR #355, opened 2025-08-28
chore(ci): enable rhel-10, s390x, and ppc64le packit build targets

jlocash:
* Card ID: CCT-1407
* Added rhel-10 build targets
* Added targets for s390x and ppc64le architectures
* Removed rhel-9 and centos-stream-9 targets


DuckBoss (changes requested):
Some dependencies for building on s390x and ppc64le are not provided in fedora42/43 so these checks are failing. You might need to remove the fedora targets for the architectures. 

Specifically `golang-github-containerd-devel-1.6.23-1` isn't finding a provider in the logs.

It looks like rhc isn't affected by this though :thinking:

PR #356, opened 2025-09-01
test: Add integration test for local privilege escalation vulnerability

Archana-PandeyM:
Verifies that unprivileged users cannot abuse com.redhat.Yggdrasil.Dispatch() DBus method to escalate privileges via workers.

* Card id: CCT-1460

### DIFF
--- a/integration-tests/test_local_privilege_escalation.py
+++ b/integration-tests/test_local_privilege_escalation.py
@@ -1,0 +1,66 @@
+import subprocess
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def test_local_privilege_escalation():
+    """
+    :id: 09ca0098-6dcd-465b-9f10-262776b858b0
+    :title: Verify unprivileged user cannot dispatch commands to workers
+    :description:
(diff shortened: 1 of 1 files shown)


integration-tests/test_local_privilege_escalation.py
```diff
@@ -0,0 +1,66 @@
...
+        # Attempt to dispatch package installation command as non-privileged user
+        cmd = f"""echo '{{"command":"install","name":"{package_name}"}}' | yggctl dispatch --worker package_manager -"""
+        logger.info(f"Attempting package installation as unprivileged user {test_user}: {cmd}")
+        result_login = subprocess.run(
+            ["su", "-", test_user, "-c", cmd],
+            capture_output=True,
+            text=True,
+        )
+
+        assert result_login.returncode == 1
```
Lorquas, line 58: This should probably be `!= 0` in case this ever changes (and its what we actually care about)

Lorquas (changes requested):
LGTM, small changes.

PR #392, opened 2026-03-30
fix: set HTTP client timeout and max idle connection timeout

pablomh:
## Problem

When yggdrasil is used for MQTT-based Remote Execution (ReX) in
Foreman/Satellite, jobs frequently get stuck in **Pending** state even
though the host has successfully executed the job.

Investigation of the yggdrasil logs on an affected host revealed the
following pattern:

1. The host receives a job via MQTT and fetches its details with a `GET`
   to the smart proxy — this succeeds immediately (HTTP 200).
2. The host executes the job (e.g. a `dnf` package install taking
   several minutes).
3. The host sends a first `POST .../update` to report progress — this
   succeeds (HTTP 200).
4. Some time later (> 30 seconds after the previous request), the host
   sends a second `POST .../update` to report completion — **no response
   is ever logged, and yggd hangs indefinitely**.

Cross-referencing with the smart proxy access log confirmed that **the
second POST never arrived at the server**. The proxy logged a `Started`
and `Finished` entry for every request it received — the stuck POST left
no trace on the server side.

## Root cause

`foreman-proxy` uses WEBrick as its HTTP server, which has a default
`KeepAliveTimeout` of **30 seconds**. After 30 seconds of inactivity on
a keep-alive connection, WEBrick silently closes its end of the TCP
socket.

`NewHTTPClient` clones `http.DefaultTransport`, which has an
`IdleConnTimeout` of **90 seconds**. This means yggd considers
keep-alive connections valid for up to 90 seconds, well past the point
where the server has already closed them.

When the gap between two consecutive requests on the same connection
exceeds the server's keep-alive timeout, yggd writes into a dead socket.
In environments where the TCP RST/FIN from the server is dropped by an
intermediate network element (e.g. a NAT or container network bridge),
the Go HTTP client has no way to detect the dead connection and blocks
forever waiting for a response — because `http.Client.Timeout` is unset
(zero = no timeout).

This bug is specific to MQTT-based ReX because SSH-based ReX has the
proxy initiate the connection to the host directly; the host never needs
to POST results back. Only the pull-based (MQTT) flow requires the host
to make outbound HTTP calls to the proxy, exposing this failure mode.

## Fix

- Set `IdleConnTimeout` on the transport so that idle connections are
  proactively closed on the client side before the server's keep-alive
  timeout fires, avoiding the stale connection race.
- Set `Timeout` on the `http.Client` as a safety net: if a stale
  connection is reused anyway and the server never responds, the request
  fails with a timeout error instead of blocking indefinitely.

Both values are exposed as configurable options — `http-idle-conn-timeout`
and `http-timeout` respectively — following the existing convention for
HTTP and MQTT tuning knobs. `http-idle-conn-timeout` defaults to 30s.
`SetIdleConnTimeout` is introduced as a method on `Client` to allow
callers to apply the configured value without leaking transport
internals outside the package.

While the specific keep-alive timeout value of 30 seconds was observed
with WEBrick/foreman-proxy, the fix is not specific to that server —
setting a client-side idle timeout and a request timeout is correct
practice for any HTTP client.

### DIFF
--- a/cmd/yggd/main.go
+++ b/cmd/yggd/main.go
@@ -53,25 +53,27 @@ func generateDocumentation(c *cli.Context) error {
 // CLI flags and arguments.
 func setupDefaultConfig(c *cli.Context) {
 	config.DefaultConfig = config.Config{
-		LogLevel:                 c.String(config.FlagNameLogLevel),
-		ClientID:                 c.String(config.FlagNameClientID),
-		Server:                   c.StringSlice(config.FlagNameServer),
-		CertFile:                 c.String(config.FlagNameCertFile),
-		KeyFile:                  c.String(config.FlagNameKeyFile),
-		CARoot:                   c.StringSlice(config.FlagNameCaRoot),
-		PathPrefix:               c.String(config.FlagNamePathPrefix),
-		Protocol:                 c.String(config.FlagNameProtocol),
-		DataHost:                 c.String(config.FlagNameDataHost),
-		FactsFile:                c.String(config.FlagNameFactsFile),
-		HTTPRetries:              c.Int(config.FlagNameHTTPRetries),
-		HTTPTimeout:              c.Duration(config.FlagNameHTTPTimeout),
-		MQTTConnectRetry:         c.Bool(config.FlagNameMQTTConnectRetry),
-		MQTTConnectRetryInterval: c.Duration(config.FlagNameMQTTConnectRetryInterval),
-		MQTTAutoReconnect:        c.Bool(config.FlagNameMQTTAutoReconnect),
-		MQTTReconnectDelay:       c.Duration(config.FlagNameMQTTReconnectDelay),
-		MQTTConnectTimeout:       c.Duration(config.FlagNameMQTTConnectTimeout),
-		MQTTPublishTimeout:       c.Duration(config.FlagNameMQTTPublishTimeout),
-		MessageJournal:           c.String(config.FlagNameMessageJournal),
+		LogLevel:                  c.String(config.FlagNameLogLevel),
+		ClientID:                  c.String(config.FlagNameClientID),
+		Server:                    c.StringSlice(config.FlagNameServer),
+		CertFile:                  c.String(config.FlagNameCertFile),
+		KeyFile:                   c.String(config.FlagNameKeyFile),
+		CARoot:                    c.StringSlice(config.FlagNameCaRoot),
+		PathPrefix:                c.String(config.FlagNamePathPrefix),
+		Protocol:                  c.String(config.FlagNameProtocol),
+		DataHost:                  c.String(config.FlagNameDataHost),
+		FactsFile:                 c.String(config.FlagNameFactsFile),
+		HTTPRetries:               c.Int(config.FlagNameHTTPRetries),
+		HTTPTimeout:               c.Duration(config.FlagNameHTTPTimeout),
+		HTTPIdleConnTimeout:       c.Duration(config.FlagNameHTTPIdleConnTimeout),
+		HTTPResponseHeaderTimeout: c.Duration(config.FlagNameHTTPResponseHeaderTimeout),
+		MQTTConnectRetry:          c.Bool(config.FlagNameMQTTConnectRetry),
+		MQTTConnectRetryInterval:  c.Duration(config.FlagNameMQTTConnectRetryInterval),
+		MQTTAutoReconnect:         c.Bool(config.FlagNameMQTTAutoReconnect),
+		MQTTReconnectDelay:        c.Duration(config.FlagNameMQTTReconnectDelay),
+		MQTTConnectTimeout:        c.Duration(config.FlagNameMQTTConnectTimeout),
+		MQTTPublishTimeout:        c.Duration(config.FlagNameMQTTPublishTimeout),
+		MessageJournal:            c.String(config.FlagNameMessageJournal),
 	}
 }
 
@@ -175,6 +177,7 @@ func setupClient(
 			tlsConfig,
 			UserAgent,
 			time.Second*5,
+			config.DefaultConfig.HTTPResponseHeaderTimeout,
 		)
 		if err != nil {
 			return nil, nil, cli.Exit(fmt.Errorf("cannot create HTTP transport: %w", err), 1)
@@ -227,6 +230,17 @@ func setupMessageJournal(client *Client) error {
 	return nil
 }
 
+// configuredHTTPClient creates an HTTP client with all config-driven settings
+// applied (retries, timeouts).
+func configuredHTTPClient(tlsConfig *tls.Config) *http.Client {
+	client := http.NewHTTPClient(tlsConfig, UserAgent)
+	client.Retries = config.DefaultConfig.HTTPRetries
+	client.Timeout = config.DefaultConfig.HTTPTimeout
+	client.SetIdleConnTimeout(config.DefaultConfig.HTTPIdleConnTimeout)
+	client.SetResponseHeaderTimeout(config.DefaultConfig.HTTPResponseHeaderTimeout)
+	return client
+}
+
 // setupTLS tries to set up new TLS config and HTTP client
 func setupTLS() (*http.Client, *tls.Config, error) {
 	tlsConfig, err := config.DefaultConfig.CreateTLSConfig()
@@ -234,11 +248,7 @@ func setupTLS() (*http.Client, *tls.Config, error) {
 		return nil, nil, cli.Exit(fmt.Errorf("cannot create TLS config: %w", err), 1)
 	}
 
-	httpClient := http.NewHTTPClient(tlsConfig, UserAgent)
-	httpClient.Retries = config.DefaultConfig.HTTPRetries
-	httpClient.Timeout = config.DefaultConfig.HTTPTimeout
-
-	return httpClient, tlsConfig, nil
+	return configuredHTTPClient(tlsConfig), tlsConfig, nil
 }
 
 // publishConnectionStatus tries to publish connection status to server
@@ -331,8 +341,7 @@ func monitorCertificate(
 		log.Info("transport TLS configuration reloaded")
 
 		log.Debug("setting dispatcher HTTP client")
-		httpClient := http.NewHTTPClient(cfg, UserAgent)
-		dispatcher.HTTPClient = httpClient
+		dispatcher.HTTPClient = configuredHTTPClient(cfg)
 		log.Info("dispatcher HTTP client updated")
 	}
 }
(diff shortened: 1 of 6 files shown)


cmd/yggd/main.go
```diff
@@ -237,6 +238,7 @@ func setupTLS() (*http.Client, *tls.Config, error) {
 	httpClient := http.NewHTTPClient(tlsConfig, UserAgent)
 	httpClient.Retries = config.DefaultConfig.HTTPRetries
 	httpClient.Timeout = config.DefaultConfig.HTTPTimeout
```
jirihnidek: Note: The timeout of http client is already set here. The value can be configured in `/etc/yggdrasil/config.toml` file using configuration option `http-timeout` or using hidden CLI option `--http-timeout`.

internal/http/client.go
```diff
@@ -26,10 +26,14 @@ type Client struct {
...
 // user-agent string.
 func NewHTTPClient(config *tls.Config, ua string) *Client {
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.TLSClientConfig = config.Clone()
+	transport.IdleConnTimeout = 30 * time.Second
+
 	client := http.Client{
-		Transport: http.DefaultTransport.(*http.Transport).Clone(),
+		Transport: transport,
+		Timeout:   30 * time.Second,
```
jirihnidek: This is IMHO also not needed. If you want to set a new default value of http client timeout, then please do it in `yggd/main.go` in `main()` function:

```go
altsrc.NewDurationFlag(&cli.DurationFlag{
    Name:   config.FlagNameHTTPTimeout,
    Usage:  "Wait for `DURATION` before cancelling an HTTP request",
    Value:  30 * time.Second,
    Hidden: true,
}),
```

internal/http/client.go
```diff
@@ -26,10 +26,14 @@ type Client struct {
 // NewHTTPClient creates a client with the given TLS configuration and
 // user-agent string.
 func NewHTTPClient(config *tls.Config, ua string) *Client {
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.TLSClientConfig = config.Clone()
+	transport.IdleConnTimeout = 30 * time.Second
```
jirihnidek: This is IMHO not needed. You set the default value in the `yggd/main.go` in the `main()` method. It is not needed to set this default value again here.

jirihnidek (changes requested):
Hi, and thank you for your PR. I have a few comments and requests.

BTW: If you want to check that values are set correctly, then you can add following code to `setupTLS()` function in `yggd/main.go`:

```go
import (
    _http "net/http"
)

log.Debugf("created HTTP client with TLS config: %+v", tlsConfig)
log.Debugf("created HTTP client with retries: %v", httpClient.Retries)
log.Debugf("created HTTP client with timeout: %v", httpClient.Timeout)
idleTimeout := httpClient.Transport.(*_http.Transport).IdleConnTimeout
log.Debugf("created HTTP client with idle connection timeout: %v",
     idleTimeout)
```

pablomh:
Thanks for the feedback — updated the PR to move defaults to the flag `Value` fields instead of hardcoding them in `NewHTTPClient`.

One thing worth flagging: two other call sites create HTTP clients without applying config values — `monitorCertificate()` and `NewHTTPTransport()`. With the defaults removed from `NewHTTPClient`, those clients will fall back to Go's defaults (no timeout, `IdleConnTimeout` of 90s), which could expose the same hang on those paths.

Is this an intentional trade-off, or should those call sites also be wired up to the config? Happy to go either way.

cmd/yggd/main.go
pablomh: `http-timeout` and `http-idle-conn-timeout` control different phases of the connection lifecycle and need to be tuned independently:

- `http-timeout` is a **per-request deadline** — how long a single HTTP exchange (connect + send + receive) is allowed to take. It should be set based on expected server response times.
- `http-idle-conn-timeout` is a **connection pool lifetime** — how long an idle keep-alive connection can sit unused before being discarded client-side. It must be set to a value *shorter than the server's keep-alive timeout* to prevent reusing connections the server has already closed.

These two constraints do not move together: a deployment with a slow server might need a long `http-timeout` but still require a short `http-idle-conn-timeout` to match an aggressive server-side keep-alive setting. Reusing `http-timeout` for both would also silently re-introduce the bug when `http-timeout` is left at its zero default (no limit), since that would set `IdleConnTimeout` to zero as well.

jirihnidek: I know. It was just a note.

jirihnidek (dismissed):
Thanks for your update. LGTM

pbartiko (changes requested):
Hi, thanks for the work on this PR!

I've been testing it on RHEL 10 using a mock HTTPS server to simulate the WEBrick keep-alive drop behavior. The mock server is set up to accept the TLS handshake and the first request, but then it silently ignores the second request on the same socket without sending any headers or a FIN packet.

During testing, I noticed that the `http-timeout` and `http-idle-conn-timeout` changes introduced here don't actually prevent the deadlock. The `GET /control/in` thread still ends up hanging indefinitely while waiting for a response on the stale keep-alive connection. It seems that the global `http.Client.Timeout` isn't taking effect in this specific scenario, possibly because it's being overridden by the context used for long-polling.


```
# 1. First request connects and establishes Keep-Alive, then server freezes
10:46:42 sending HTTP request: GET https://.../control/.../in
10:46:42 received HTTP response: &{200 OK ... map[Connection:[keep-alive] Content-Length:[0]] ...}

# 2. The data/in thread throws retry errors and loops properly
10:46:47 sending HTTP request: GET https://.../data/.../in
10:47:22 sending HTTP request: GET https://.../data/.../in
10:47:57 sending HTTP request: GET https://.../data/.../in

# 3. The control/in thread is never seen again (deadlocked)
```

To see what was missing, I manually modified `internal/http/client.go` to include `ResponseHeaderTimeout: 10 * time.Second` directly in the `http.Transport` configuration. Once I added that, the deadlock was successfully broken. The client correctly detected the dead connection, closed the socket, and resumed its normal retry loop:

```
# 1. First request connects and establishes Keep-Alive, then server freezes
10:57:21 sending HTTP request: GET https://.../control/.../in
10:57:21 received HTTP response: &{200 OK ... map[Connection:[keep-alive] Content-Length:[0]] ...}

# 2. SUCCESS: Both threads detect the dead connection and retry
10:57:26 sending HTTP request: GET https://.../data/.../in
10:57:36 sending HTTP request: GET https://.../control/.../in   <-- (Recovered)

10:58:01 sending HTTP request: GET https://.../data/.../in
10:58:11 sending HTTP request: GET https://.../control/.../in   <-- (Looping correctly)

10:58:36 sending HTTP request: GET https://.../data/.../in
10:58:46 sending HTTP request: GET https://.../control/.../in
```

On a positive note, I ran the existing integration tests and they all passed successfully. The current changes are safe and shouldn't disrupt any existing functionality.

pablomh:
Thanks for testing this so carefully, @pbartiko — the diagnosis is spot on.

The root cause was simpler than it looked: `NewHTTPTransport` creates its own internal HTTP client, and the previous version only wired the timeout settings to the dispatcher's client (the one doing `POST .../out`). The goroutines doing `GET /control/in` and `GET /data/in` use a completely separate client with no timeout at all — so nothing was ever going to interrupt the hang.

`ResponseHeaderTimeout` is also a better fit than `Client.Timeout` for the long-polling path specifically: `Client.Timeout` covers the entire request lifecycle including reading the full response body, which means it would also cancel legitimately held-open polling connections after 30s. `ResponseHeaderTimeout` only times out the wait for the first response byte, so live connections aren't affected but frozen ones get cut quickly. The dispatcher's client gets both as defence-in-depth (its requests always complete fully), the transporter's client only gets `ResponseHeaderTimeout`.

To validate the fix we built a local HTTPS mock server that reproduces your setup — accepts the TLS handshake, responds 200 to the first request on a connection, then freezes on the second with no response and no FIN. We confirmed the hang before the fix and that the connection is correctly detected and retried after it.

**What changed**

- New `http-response-header-timeout` flag (default 30s, hidden, same pattern as the other knobs).
- `SetResponseHeaderTimeout` on `internal/http.Client`, same shape as `SetIdleConnTimeout`.
- `NewHTTPTransport` now takes `responseHeaderTimeout` and applies it to its internal client, covering the long-polling path. Stored on the struct so `ReloadTLSConfig` re-applies it when recreating the client on cert rotation.
- Extracted `configuredHTTPClient()` helper so initial setup and cert-reload both apply the same set of settings.
- Tests for `SetIdleConnTimeout`/`SetResponseHeaderTimeout` and a behavioral test that verifies the timeout survives a `ReloadTLSConfig` call.

**One question on the tests**

The `ReloadTLSConfig` test uses `httptest.NewServer` to spin up a per-test slow server inline, rather than extending the existing `startServer()` helper. The reason is that `startServer()` registers handlers on the global `net/http` default mux, so adding a slow handler there would affect all tests. Happy to go either way — is `httptest.NewServer` fine here, or would you prefer keeping a single server setup for consistency?

pbartiko:
That's a great update! Thanks for the detailed explanation, that makes a lot of sense. It perfectly explains why we saw the dispatcher recover while the polling threads were still hanging. I'm really glad the mock server approach helped reproduce the issue on your end too.

I just pulled the latest changes and tested them locally with the mock server and the new `--http-response-header-timeout 10s` flag. Everything seems to work great now - both threads successfully detect the dead connection and recover.

```
# 1. First request connects and establishes Keep-Alive, then server freezes
06:51:15 sending HTTP request: GET https://.../control/.../in
06:51:15 received HTTP response: &{200 OK ... map[Connection:[keep-alive] Content-Length:[0]] ...}

# 2. SUCCESS: Both threads detect the dead connection and retry
06:51:20 sending HTTP request: GET https://.../data/.../in
06:51:30 sending HTTP request: GET https://.../control/.../in   <-- (Recovered)

06:51:55 sending HTTP request: GET https://.../data/.../in
06:52:05 sending HTTP request: GET https://.../control/.../in   <-- (Looping correctly)
```

Regarding your question about the tests, I personally lean towards keeping `httptest.NewServer` for this specific case. Since `startServer()` uses the global default mux, I'd be a bit worried that adding a slow handler there might accidentally affect other tests running at the same time and make them flaky. Using an isolated server just for this timeout check feels a bit safer to me, but I'm completely fine with whichever approach you prefer for the project.

Thanks again for the quick turnaround and all the effort you put into fixing this.

pbartiko (approved):

pablomh:
Glad the fix works on your end too, thanks for re-testing so quickly!

And yes, that's exactly why we went with `httptest.NewServer` — same concern about the global mux making other tests flaky.

jirihnidek (approved):
LGTM

BTW: I found some other issue, which is relate to workers working in the remote mode. I will create another Jira card for this issue.

pablomh:
@jirihnidek, would you mind mentioning it here once you have it ready so I can keep track of it? Thanks for your time!!

PR #395, opened 2026-04-15
fix: (0.2.1) Update grpc to version 1.79.3

jirihnidek:
* Card ID: (CCT-2235)[https://redhat.atlassian.net/browse/CCT-2235]
* `grpc` 1.79.3 requires newer version of Go (1.24)
* The `PACKAGE` was not used. It was probably typo
* Updated GitHub actions


go.mod
```diff
@@ -1,22 +1,42 @@
 module github.com/redhatinsights/yggdrasil
 
-go 1.13
+go 1.25.0
```
DuckBoss, line 3: go 1.24.0 seems to be the minimum version for grpc 1.79.3, why do we need go 1.25.0?

jirihnidek: I did not create this manually. Tools updated the version according some other dependencies. BTW: RHEL 8.6 contains golang 1.25. We should be good.

DuckBoss (dismissed):

go.mod
jirihnidek: I run it once again and it was not so aggressive in update. It requires version 1.24 too.

.github/workflows/go.yml
```diff
@@ -10,15 +10,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ['1.15', '1.16', '1.17']
+        go: ['1.25']
```
DuckBoss, line 13: @jirihnidek With the go version being bumped down to 1.24.0 in your recent changes, this workflow matrix should probably be bumped down too.

go.mod
```diff
@@ -1,22 +1,40 @@
 module github.com/redhatinsights/yggdrasil
 
-go 1.13
+go 1.24.0
```
DuckBoss, line 3: Your commits still refer to go being updated to `1.25` instead of `1.24`.
Can you correct these descriptions/titles?

DuckBoss (changes requested):

.github/workflows/go.yml
jirihnidek: The RHEL-8 contains Go 1.25. For this reason it makes sense to test it with Go 1.25 in GitHub action too.

go.mod
jirihnidek: Commit log was updated.

jirihnidek:
There is not required Go version on rhel-8.6. Closing.

PR #399, opened 2026-04-16
fix: (0.2.6) Update grpc to version 1.79.3

jirihnidek:
* Card ID: CCT-2235
* `grpc` 1.79.3 requires newer version of Go (1.24)
* Updated `gopkg.in/yaml.v2` to version `2.2.8` without CVE


DuckBoss (approved):
LGTM, thanks!

PR #400, opened 2026-04-22
fix: set HTTP client timeouts to prevent indefinite hangs

pablomh:
## Problem

When yggdrasil is used for MQTT-based Remote Execution (ReX) in
Foreman/Satellite, jobs can get stuck in **Pending** state because
yggd hangs indefinitely while trying to POST job results back to
the smart proxy.

The root cause is the same as #392 (fixed in main): `initHTTPClient`
creates an `http.Client` with no timeouts. When the server closes a
keep-alive connection after its keep-alive timeout fires, and the
TCP FIN/RST is dropped by an intermediate network element (e.g. a
NAT or container network bridge), the client has no way to detect
the dead connection and blocks forever.

## Fix

Set three timeouts in `initHTTPClient`:

- `IdleConnTimeout` (30s): proactively close idle connections before
  the server's keep-alive timeout fires, reducing the chance of
  reusing a stale connection.
- `ResponseHeaderTimeout` (30s): fail quickly if the server never
  sends response headers, rather than blocking indefinitely.
- `Timeout` (30s): safety net covering the full request lifecycle.

Unlike the main branch, all three timeouts are safe to set here
because the HTTP client in 0.2.x is only used for normal
request/response cycles (fetching detached job content and posting
results). The control plane is MQTT, so there is no long-polling
over HTTP to worry about.

## Notes

This is a backport of #392 to the 0.2.x series. The fix is simpler
here since there is a single shared HTTP client with one construction
site and no cert-reload path to worry about.

### DIFF
--- a/cmd/yggd/http.go
+++ b/cmd/yggd/http.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/redhatinsights/yggdrasil"
 	"github.com/subpop/go-log"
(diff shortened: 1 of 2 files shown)


jirihnidek (approved):
Thanks for the PR. LGTM

pablomh:
Ping?

PR #401, opened 2026-04-29
test: mark integration tests with pytest tier1

pbartiko:
Register the `tier1` marker in `integration-tests/conftest.py` so pytest does not warn on unknown marks, and apply `@pytest.mark.tier1` to all integration tests.


Aftermath (approved):
LGTM.

PR #406, opened 2026-06-17
docs: add AGENTS.md for AI-assisted development

pbartiko:
* Card ID: CCT-2592
* Document repository layout, build (Go and meson), and testing workflows
* Describe CI checks, security expectations, and contribution conventions
* Point agents to CONTRIBUTING.md, README.md, and worker/echo for detail

Assisted-by: Cursor

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,224 @@
+# AGENTS.md
+
+This file contains guidance for coding agents when helping with development in
+this repository.
+
+## What is included in this repository
+
+This repository contains source code for two binaries: `yggd`, which runs as a
+system daemon, and `yggctl`, which is a CLI tool for controlling a running
+`yggd` instance. It also includes an example worker, [worker/echo](worker/echo),
+showing how to implement the `com.redhat.Yggdrasil1.Worker1` interface.
+
+_yggd_ subscribes to topics on an MQTT broker and routes data to child worker
+processes over D-Bus. The dispatcher exposes `com.redhat.Yggdrasil1.Dispatcher1`.
+
+Release lines in this repository: **`main`** (0.4.x, RHEL 10+) and maintenance branches
+such as **`yggdrasil-0.2.x`** (RHEL 9 era). Distro package names do not always
+match the branch — on RHEL 10+, the daemon ships in the **`yggdrasil`** RPM
+(0.4.x); on RHEL 9, the equivalent **`rhcd`** service was provided through the
+**`rhc`** RPM (0.2.x source line). The [rhc](https://github.com/RedHatInsights/rhc)
+project is a separate repository — yggdrasil source code has never been part of
+that repo. `rhc` still activates `yggd`/`rhcd` when connecting a host to Red Hat
+services.
+
+| Path | Purpose |
+|------|---------|
+| `cmd/yggd` | Main system daemon |
+| `cmd/yggctl` | Control CLI for `yggd` (always installed to `bindir` by meson/RPM, e.g. `/usr/bin/yggctl`) |
+| `internal/` | Private packages (`config`, `transport`, `work`, `messagejournal`, …) |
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,197 @@
+# AGENTS.md
+
+This file contains guidance for coding agents when helping with development in
+this repository.
+
+## What is included in this repository
+
+_yggdrasil_ is a system daemon (`yggd`) that subscribes to topics on an MQTT
```
jirihnidek: I would start with statement that this repository contains source code for two binaries: `yggd`, which works as a daemon and  `yggctl`, which works as CLI tool for the `yggd` daemon.

jirihnidek: And I would also say that this repository contains example of the worker called `echo` worker.

AGENTS.md
```diff
@@ -0,0 +1,197 @@
...
+broker and routes data to child "worker" processes over D-Bus. Workers implement
+the `com.redhat.Yggdrasil1.Worker1` interface; the dispatcher exposes
+`com.redhat.Yggdrasil1.Dispatcher1`.
+
+On RHEL 10+, yggdrasil ships as a standalone RPM. On earlier RHEL releases it
```
jirihnidek: It would be important to say what version of yggdrasil is shipped on RHEL 10.

It is also important to say what version of yggdrasil was bundled as part of `rhc` RPM.

AGENTS.md
```diff
@@ -0,0 +1,197 @@
...
+was bundled inside [rhc](https://github.com/RedHatInsights/rhc); `rhc` still has
```
jirihnidek: I understand you what you want to say, but it is not accurate. It could be misleading information, because `yggdrasil` source code has never been part of this `rhc` repository.

AGENTS.md
```diff
@@ -0,0 +1,197 @@
...
+
+Go unit tests cover library packages under `internal/` and selected top-level
+code. End-to-end MQTT and D-Bus behavior is exercised by
+`integration-tests/` (Python), not by Go unit tests alone.
+
+### Unit tests (Go)
+
+```bash
+go test -v ./...
+go vet -v ./...
```
jirihnidek, line 102: The `go vet` does not run unit test. It runs various static analyzers and it checks correctness of source code.

AGENTS.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+CI also runs **commitsar** (conventional commits), **woke** (inclusive language),
+and **Packit** (COPR build + TMT system tests). Lint steps are defined in
+[`.github/workflows/lint.yml`](.github/workflows/lint.yml).
+
+### Integration tests (Python)
+
+Requires mosquitto, D-Bus, and systemd. See [CONTRIBUTING.md](CONTRIBUTING.md)
+for prerequisites. Start the broker before running tests:
```
jirihnidek: I think that it is important to emphasize here that `yggdrasil` has to be installed in the system from the RPM to be able to run integration tests.

AGENTS.md
```diff
@@ -0,0 +1,197 @@
...
+
+### System tests (TMT / Testing Farm)
+
+In-tree system tests: [systemtest/plans/main.fmf](systemtest/plans/main.fmf).
+Triggered on PRs via Packit (`.packit.yaml`, label `unit`) and optionally via
+Testing Farm (`.testing-farm.yaml`).
+
+## Security
+
+`yggd` runs as a **privileged system daemon** with D-Bus name ownership and
```
jirihnidek: The `yggd` is run as system user. It has only privileges to dispatch messages to worker. The system user is part of `rhsm` group to be able to read consumer cert and key.

jirihnidek (changes requested):
Thanks for the PR. I have few comments and requests.

jirihnidek (approved):
LGTM

PR #410, opened 2026-06-24
fix(deps): updated golang.org/x/net to v0.55.0

DuckBoss:
* golang.org/x/net v0.55.0 requires an updated version of Go (1.25.0)
* golang.org/x/sys indirect dependency updated to v0.45.0


jirihnidek:
@DuckBoss It LGTM, there is only one thing that concerns me and it is generic thing. These failed builds for Fedora. I think that we should use same approach as `rhc` uses.

jirihnidek (approved):
LGTM

